PR #3, opened 2021-12-03
Add "vertical layout" template

cavasinf:
## Description
Add the "vertical" Tabler layout to the bundle.

Tabler demo link : https://preview.tabler.io/layout-vertical.html

## Types of changes

## Checklist


cavasinf:
As I can see, `<head>`, `.navbar-nav` and `.page-wrapper` are the same between horizontal layout and vertical layout.

If you think this template is correct, 
I can't try to make a `_base-layout` template instead of duplicates parts in common in `vertical` and `horizontal` layout.

kevinpapst:
Can you share a screenshot - what is a vertical layout?

cavasinf:
@kevinpapst Demo links from tabler.io and screnshot from your bundle.

### Vertical menu 
- Demo link: https://preview.tabler.io/layout-vertical.html
- How it looks with your bundle: (navbar_condensed : true && user_menu_condensed : false)
image


### Vertical menu with navbar 
- Demo link: https://preview.tabler.io/layout-combo.html
- How it looks with your bundle: (navbar_condensed : false)
image



Here's a responsive menu:
image

kevinpapst:
Aha, understand. Yeah, let's try to make it one file. Otherwise changes need to be carried over, what will be forgotten at some point...

cavasinf:
There is a transparent theme too: https://preview.tabler.io/layout-vertical-transparent.html

But there's no variable in `tabler.yaml` to configure that mode (true/false).
I don't want to break everything right now OR make it complicated.

cavasinf:
> Aha, understand. Yeah, let's try to make it one file. Otherwise changes need to be carried over, what will be forgotten at some point...

OK, I'll make it into one file.
I'll commit in the same branch, so same merge request

cavasinf:
- `tabler.yaml` Added new **required** [enum] option `layout`
  - `horizontal` (default)
  - `vertical`
- Pooling "horizontal" and "vertical" layout into `layout.html.twig`
- Didn't touch at `RuntimeExtensionTest.php`

cavasinf:
❔ Questions ❔ 

1. Layouts are into `templates/includes/layout/`, did you prefer in another folder/location?

kevinpapst:
I don't think we will end up with dozens of includes, so `templates/includes/layout_horizontal.html.twig` would be enough for me.

kevinpapst:
General feedback: I know the wording is based on Tabler demo, but I'd prefer `horizontal vs vertical` "navigation" and not "layout".

src/Helper/ContextHelper.php
```diff
@@ -66,6 +66,26 @@ public function isBoxedLayout(): bool
         return (bool) $this->getOption('boxed_layout');
     }
 
+    public function setLayout(string $layout): void
```
kevinpapst, line 69: Regardless of the layout vs menu wording question, I would prefer fail-safe methods `activateHorizontaMenu()` vs `activateVerticalMenu()` or at least `setIsHorizontalMenu(bool $isHorizontalMenu)`.

src/Helper/ContextHelper.php
```diff
@@ -66,6 +66,26 @@ public function isBoxedLayout(): bool
...
+
+    public function isLayoutHorizontal(): bool
+    {
+        return $this->getOption('layout') === "horizontal";
+    }
+
+    public function isLayoutVertical(): bool
+    {
+        return $this->getOption('layout') === "vertical";
+    }
```
kevinpapst, line 87:
```suggestion
    public function isLayoutVertical(): bool
    {
        return !$this->isLayoutHorizontal();
    }
```

src/Helper/ContextHelper.php
```diff
@@ -66,6 +66,26 @@ public function isBoxedLayout(): bool
...
+
+    public function getLayout(): string
+    {
+        return (string) $this->getOption('layout');
+    }
+
+    public function isLayoutHorizontal(): bool
+    {
+        return $this->getOption('layout') === "horizontal";
+    }
```
kevinpapst, line 82:
```suggestion
    public function isLayoutHorizontal(): bool
    {
        return $this->getLayout() === 'horizontal';
    }
```

templates/includes/menu.html.twig
```diff
@@ -2,7 +2,7 @@
     {% set menu = tabler_menu(app.request) %}
 {% endif %}
 {% if menu is defined and menu is not null %}
-<ul class="navbar-nav">
+<ul class="navbar-nav {{ navbar_class | default('') }}">
```
kevinpapst, line 5: What is `navbar_class`?

cavasinf: Used to add padding to when in vertical mode.
https://github.com/cavasinf/TablerBundle/blob/b4d91a7721bfe22fffb3e5d087b21b6aee64cf2a/templates/includes/layout/vertical.html.twig#L22

src/Helper/ContextHelper.php
cavasinf: It has been done in that way to predict the `overlap` layout:
https://preview.tabler.io/layout-navbar-overlap.html

In that case it is not only the "menu" that change, but also the `page-wrapper`.
At the end, there's not only two mode.

How do you see a better solution ?

src/Helper/ContextHelper.php
```diff
@@ -66,6 +66,26 @@ public function isBoxedLayout(): bool
         return (bool) $this->getOption('boxed_layout');
     }
 
+    public function setLayout(string $layout): void
+    {
+        $this->setOption('layout', $layout);
+    }
+
+    public function getLayout(): string
```
kevinpapst, line 74: Thinking about it again: this could be private, right?

templates/includes/menu.html.twig
cavasinf: Refer to https://github.com/tabler/tabler/blob/1018977e7f9d0d09c8e450b15b313b5c2d7d0524/demo/layout-vertical-right.html#L83

src/Helper/ContextHelper.php
kevinpapst: What speaks against a third `activateOverlapMenu()` ? I'd try to avoid leaking internal information. How we store the layout information in that helper shouldn't be a problem of the user (dev).

src/Helper/ContextHelper.php
cavasinf: Yes it could.
If you don't want to allow the user (of the bundle) get the real value behind `layout` option.

src/Helper/ContextHelper.php
cavasinf: Overlap layout look like horizontal layout but in code it is not.
It overlap (lol) inside the `page-wrapper`. 

By that I mean, the way of thinking for us (bundle devs) is not simply a vertical layout WITH an overlap mode.
It should be new/different layout Overlap.

src/Helper/ContextHelper.php
kevinpapst: Once you expose it, you have to make sure it works in the long run, which puts a higher maintenance burden on the maintainer => so using `final` and `private` wherever possible is the way to go. I don't see any value in leaking this information to the outside world.

cavasinf: Ok I get it now. I'll make it private
Getter/Setter or only Getter ?

kevinpapst: The getter definitely and we keep on discussing about the setter 😁

cavasinf: Updated with ccc4f46

src/Helper/ContextHelper.php
cavasinf: Still trying to figure it out what is the best way..
I can't get over a third layout.

## Questioning 
With what we are exchanging: 
- Layout Vertical
- Layout Horizontal 
    -> Overlap

If we are going in that way, dev can activate option `overlap` from tabler.yaml but without any changes when vertical  layout is defined. 
It could be disturbing and/or weird to have access to an option in only certain case/mode.

## Suggest
The best way for me is:
- Layout Vertical
- Layout Horizontal
- Layout Overlap (Horizontal)

## For the future
At the end, 
if we are including all templates that Tabler provides, 
we will still have options available in certain case Like "Transparent" only linked to "Vertical Layout".

If I try to visualize it as "TreeBuilder", it can look like:
- Layout Vertical
    -> Transparent [option of vertical]
  => Right
- Layout Horizontal
    -> Dark [option of horizontal]
  => Sticky
  => Condensed
    -> Dark [option of horizontal -> condensed]
- Layout Overlap (Horizontal like)

In that way, Overlap has no option available which is the case in Tabler

templates/layout.html.twig
```diff
@@ -24,9 +24,11 @@ Enjoy your theme!
 <div class="wrapper">
 
     {% if tabler_bundle.isLayoutHorizontal() %}
-        {% include '@Tabler/includes/layout_horizontal.html.twig' %}
+        {% use '@Tabler/includes/layout_horizontal.html.twig' %}
+        {% block layout_horizontal %}{{ parent() }}{% endblock %}
     {% elseif tabler_bundle.isLayoutVertical() %}
-        {% include '@Tabler/includes/layout_vertical.html.twig' %}
+        {% use '@Tabler/includes/layout_vertical.html.twig' %}
+        {% block layout_vertical %}{{ parent() }}{% endblock %}
```
cavasinf: Changing `include` to `use`.

It allow devs to override blocks in `layout_horizontal` or `layout_vertical`.

#### Exemple
```twig
{% extends '@Tabler/layout.html.twig' %}

{% block navbar_search %}{# Disable search bar #}{% endblock %}
...
...
...
```

kevinpapst: Sure this works with the if?  Afaik the latest defined block wins and (from here I am unsure) the use should be evaluated during compile time, while the if is executed during runtime.

cavasinf: I mean, I've tried both layout and it works with both of them.

The thing that I'm not sure that is working is the auto-completion/detection.

What IDE are you using ? Mine is PHPStorm and I don't know why but your Bundle with Tabler is not recognized by it.
AdminLTEBundle work like a charm, but not that one. 
How do you declare the `@Tabler` isn't it from the `xxxxBundle.php`

TablerBundle:
image

AdminLTEBundle:
image

cavasinf: @kevinpapst I've opened an issue about that: https://github.com/kevinpapst/TablerBundle/issues/7
If you can paste your reply there. I don't want to poluate  this review with IDE configs.

cavasinf:
❗ Note/Warning ❗ 
Do not add `before` or `after` twig blocks to `header `, `nav-bar`, `aside`, ...

The CSS from Tabler.io uses tilde `~` to define styles.
See: https://github.com/tabler/tabler/blob/02c45fc09d020dd424f526384163ab01ddb58b8f/dist/css/tabler.css#L8009

When adding HTML elements, we can brake those selectors.

cavasinf:
@kevinpapst this PR is pretty BIG now.

Do you want me to do https://github.com/kevinpapst/TablerBundle/issues/43 there.
Or we merge this PR and I create another PR for that functionality ?

kevinpapst:
Do it in here and I will test and merge this one during the Christmas holidays.

cavasinf:
@kevinpapst Just added the new color param.

templates/layout.html.twig
```diff
@@ -22,64 +22,21 @@ Enjoy your theme!
...
-                        {% include '@Tabler/includes/navbar_search.html.twig' %}
-                    {% endblock %}
-                </div>
-            </div>
-        </div>
-    </div>
+
+    {% if tabler_bundle.isLayoutHorizontal() %}
+        {% use '@Tabler/includes/layout_horizontal.html.twig' %}
+        {% block layout_horizontal %}{{ parent() }}{% endblock %}
```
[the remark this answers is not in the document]
cavasinf: Hacky way to allow IDE to know the block inside the template.

Include (layer) of an include (layout_horizontal) doesn't show the blocks inside `layout_horizontal`

templates/includes/layout_vertical.html.twig
```diff
@@ -0,0 +1,55 @@
+{% block layout_vertical %}
+
+{% block sidebar %}
```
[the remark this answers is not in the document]
cavasinf:
```suggestion
{% block navbar %}
```

templates/includes/layout_horizontal.html.twig
```diff
@@ -0,0 +1,63 @@
...
+        {% endblock %}
+        <div class="navbar-nav flex-row order-md-last">
+            {% block navbar_start %}{% endblock %}
+            {% block navbar_notifications %}
+                {% include '@Tabler/includes/navbar_notifications.html.twig' %}
+            {% endblock %}
+            {% block navbar_user %}
+                {% include '@Tabler/includes/navbar_user.html.twig' %}
+            {% endblock %}
+            {% block navbar_end %}{% endblock %}
```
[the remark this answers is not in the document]
cavasinf: True 👍

src/Model/MenuItemModel.php
```diff
@@ -189,6 +189,11 @@ public function getActiveChild(): ?MenuItemInterface
         return null;
     }
 
+    public function hasActiveChildren(): bool
```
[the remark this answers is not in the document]
cavasinf: @kevinpapst I don't get what you meant by that.

One item available is considered as active when it shouldn't ?

I have that case on my project and it works fine:
imageimage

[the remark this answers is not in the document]
cavasinf: Oh .. I'm dumb...

The function name is wrong, I get it ... 🦆

src/DependencyInjection/Configuration.php
```diff
@@ -125,9 +125,19 @@ private function getOptionsConfig(): NodeDefinition
         $rootNode
             ->addDefaultsIfNotSet()
             ->children()
+                ->enumNode('layout')
+                    ->values(['horizontal', 'vertical'])
+                    ->defaultValue('horizontal')
```
cavasinf, line 130: > @kevinpapst please remove that, otherwise tabler.options node is required to be configured

What is wrong with that ?

cavasinf:
> when I switch to this branch in the demo app, I see this: Bildschirmfoto 2022-01-08 um 22 57 19 while it should look like this: Bildschirmfoto 2022-01-08 um 23 00 22 is there a hidden BC break that I missed?

@kevinpapst What is the value of `navbar_condensed` ?

templates/layout.html.twig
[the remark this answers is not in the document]
cavasinf: Removed

src/DependencyInjection/Configuration.php
kevinpapst: I already removed it. It is "wrong" because:
- it has a default values
- addDefaultsIfNotSet() is set
- it was the ONLY config that had this setting, thus forcing the user to create the config/packages/tabler.yaml

The demo app works fine without the config file if `isRequired()` is removed.
If we ship good default values, user can configure the bundle, but don't have to at the beginning. 
This should make the entry barrier for new devs smaller.

cavasinf:
That:
```twig
{% use '@Tabler/includes/layout_horizontal.html.twig' %}
{% block layout_horizontal %}{{ parent() }}{% endblock %}
```
Is not that `hacky`..

With the current code wich is now a simple include:
```twig
{% include '@Tabler/includes/layout_horizontal.html.twig' %}
```

Devs cannot override the block `navbar_search` inside `layout_horizontal.html.twig`.

templates/layout.html.twig
```diff
@@ -22,64 +22,18 @@ Enjoy your theme!
...
-                </div>
-            </div>
-        </div>
-    </div>
+
+    {% if tabler_bundle.isLayoutHorizontal() %}
+        {% include '@Tabler/includes/layout_horizontal.html.twig' %}
+    {% elseif tabler_bundle.isLayoutVertical() %}
+        {% include '@Tabler/includes/layout_vertical.html.twig' %}
     {% endif %}
```
cavasinf, line 30: Include of an include 🤔 
Doesn't works => Cannot override search block

cavasinf (commented):
Review update

templates/layout.html.twig
cavasinf: See e8c7df1

kevinpapst:
I tested now for 2 hours and can't get it to work.

- `include` doesn't allow to override blocks
- `embed` doesn't allow to override blocks
- `use` with `parent()` doesn't work, because `use` is interpreted during compile time (?) and twig always uses the blocks from the last `use` (so always vertical here) 
 
see https://twig.symfony.com/doc/3.x/tags/use.html 

 > You can use as many use statements as you want in any given template. If two imported templates define the same block, the latest one wins.

Even pasting the code from the files does not work because "block is already defined".

For now I believe the only solution is to create a new layout file for vertical navigation.

cavasinf:
Works when reversing commit e8c7df1

kevinpapst:
> Works when reversing commit [e8c7df1](https://github.com/kevinpapst/TablerBundle/commit/e8c7df1de661647c4e610490a2f93e882a46ee4f)

No it doesn't because of what I wrote above

cavasinf:
> you always trigger the blocks from vertical layout. I tested that.

Yeah you right, didn't see that for horizontal..

cavasinf:
I don't really like the duplication too.

I'm trying to find a way to do the "duplicate way" but only for the layout vertical/horizontal. 
Like redefined all block inside those templates from the include.

My goal is to disable the navbar search.
I've come with something like that.

`base.html.twig`
```twig
{% extends '@Tabler/layout.html.twig' %}

{% block navbar_search %}{# Disable search bar #}{% endblock %}
```

`@Tabler/layout.html.twig`
```twig
...
{% if block('navbar_search') is defined %}
    {% set navbar_search = block('navbar_search') %}
{% endif %}
{% embed '@Tabler/includes/layout_vertical.html.twig' %}
    {% block navbar_search %}{{ navbar_search ?? parent() }}{% endblock %}
{% endembed %}
...
```

IT IS UGLY, but I cannot find anything else 😢

cavasinf:
Horizontal layout:
- header 🔗 
- header_id 🔗 
- app_logo 🔗 
- navbar_menu_condensed
- navbar 🔗 
- navbar_id 🔗
- navbar_menu
- navbar_search 🔗 


Vertical layout:
- navbar 🔗 
- sidebar_id
- app_logo 🔗 
- navbar_menu_start
- navbar_menu_end
- header 🔗 
- header_id 🔗 
- navbar_search 🔗 


Yes, At least we have common blocks..

> Then it might be easier to duplicate the layout directly.

Hten we need to remove the `layout` tabler option. User will extend `layout_horizontal.html.twig` or `layout_vertical.html.twig` and not `layout.html.twig` anymore.

cavasinf:
> duplicate the layout directly

@kevinpapst Do you want me to do it ?

kevinpapst:
> Then we need to remove the layout tabler option. User will extend layout_horizontal.html.twig or layout_vertical.html.twig and not layout.html.twig anymore.

Which might be the better option anyway. 

> @kevinpapst Do you want me to do it ?

Yes please, I have no immediate use-case for the layout 🤷

cavasinf:
> Yes please, I have no immediate use-case for the layout

Ok 👍
I've many Client projects in dev right now, I'll probably do it next two weeks.

❔ Question ❔
There will be 2 "base" template now: 
-  `templates\layout-horizontal.html.twig`
-  `templates\layout-vertical.html.twig`

Do you want to: 
- remove `templates\layout.html.twig`.
- or we keep it and it only extend `templates\layout-horizontal.html.twig`.

It will bypass BC breaks on your side and we can remove `layout.html.twig` in the next major version.
WDYT ?

kevinpapst:
Make it 3 files with `layout` pointing to `layout-horizontal` and a direct warning like:
```
{% deprecated 'The "layout.html.twig" template is deprecated, use "layout-horizontal.html.twig" instead.' %}
```

templates/layout-vertical.html.twig
```diff
@@ -0,0 +1,153 @@
+{#
+Use this as your new starter template page, use it to start your new project,
+by adding this code to your own base template:
+
+{% extends '@Tabler/layout-vertical.html.twig' %}
+
+Enjoy your theme!
+#}
+
+{% set tabler_layoutType = "vertical" %}
```
cavasinf, line 10: Same case here

cavasinf:
@kevinpapst we needs to resolve/redo https://github.com/kevinpapst/TablerBundle/pull/64 here too.

cavasinf:
Merged main into this branch + conflict resolve.

### Overlap mode
How do you want to handle `overlap` and `condensed` mode ? 

They cannot be `on` at the same time.
image

Should be:
image




Simple warning in README is enough ? 🤔 
or do we check/condition from tabler Bundle config ?

kevinpapst:
I would say: that is up to the developer to configure properly: so README is enough.

Can you please [create a branch in the demo application](https://github.com/kevinpapst/TablerBundle-Demo) so I can test this ... I failed miserably last time I tried 😞

cavasinf:
Demo application doesn't works on my side (Docker environment).
And I'm not the administrator of my PC (Remote desktop), can't install symfony CLI, composer, PHP binaries, ...

I've no privilege to open a discussion in that repo, do you want me to open issue about it ?

kevinpapst:
Could you please summarize BC breaks / necessary changes in the initial post? 
I just added your branch to the demo app and some parts of the navbar disappeared (haven't checked further yet).

cavasinf:
> EDIT: the blocks in `{% embed '@Tabler/embeds/navbar.html.twig' %}{% endembed %}` cannot be overwritten anymore.

Ah.. yes ...
Same problem as here : https://github.com/kevinpapst/TablerBundle/pull/3#issuecomment-1022264943

Do we remove the include file and duplicate the code to allows override ?

kevinpapst:
Yeah, if there is no other way we have to duplicate it.

tacman:
My branch of the demo uses calvin's vertical layout:

https://github.com/tacman/TablerBundle-Demo/tree/tac

What's missing/broken that prevents this from being merged?  And is this the right place for the discussion?

cavasinf:
@tacman Only override of blocks in `'@Tabler/embeds/navbar.html.twig'` is missing.

I need to try this PR in the Tabler demo.
From memories, I don't think there's any BC breaks or other missing functionalities.

tacman:
OK, so if  I wanted to help out on this, I should fork cavisin's fork, make
changes, and submit them there?  Wouldn't it be easier if kevin created a
new branch on his repo and merged the vertical layout, instead of merging
it into main?

Sorry for the simple github questions, I think the code is almost
functional.  The next question is how it should be implemented,  a "layout"
key in the config, with the ability to override it (and store the layout in
the session?)

cavasinf:
@tacman 
> I should fork cavisin's fork, make changes, and submit them there?

Depend of what changes you want to make, tell me.

> The next question is how it should be implemented,  a "layout"
key in the config, with the ability to override it

We have already discuss that functionality with kevinpapst.
There's a big problem with twig, twig doesn't allow block override in embedded from extend.
See research here : https://github.com/kevinpapst/TablerBundle/pull/3#issuecomment-1022224913

ATE, your views should extend from the provided `layouts`

```twig
{% extends '@Tabler/layout-horizontal.html.twig' %}
```

OR

```twig
{% extends '@Tabler/layout-vertical.html.twig' %}
```
See : https://github.com/kevinpapst/TablerBundle/blob/efaf141278845dabceba6b37bde9567785be9292/docs/layout.md

--------------------

Just push what is missing in that PR, 
I've tested in the TablerBundle-Demo and everythings looks fine.

Waiting on @kevinpapst to approve.

templates/layout-vertical.html.twig
```diff
@@ -45,14 +45,7 @@ Enjoy your theme!
...
                 <div class="navbar-nav flex-row d-lg-none">
-                    {% block navbar_start %}{% endblock %}
-                    {% block navbar_notifications %}
-                        {% include '@Tabler/includes/navbar_notifications.html.twig' %}
-                    {% endblock %}
-                    {% block navbar_user %}
-                        {% include '@Tabler/includes/navbar_user.html.twig' %}
-                    {% endblock %}
-                    {% block navbar_end %}{% endblock %}
+                    {{ _self.navbar(block('navbar_start'), block('navbar_notifications'), block('navbar_user'), block('navbar_end')) }}
```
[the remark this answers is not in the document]
cavasinf: No it won't.

This result as `undefined` in the variable.

templates/layout-vertical.html.twig
```diff
@@ -165,3 +151,22 @@ Enjoy your theme!
 {% endblock %}
 </body>
 </html>
+
+{% macro navbar(navbar_start, navbar_notifications, navbar_user, navbar_end) %}
+    {% block navbar_start %}{{ navbar_start ?? '' | raw }}{% endblock %}
```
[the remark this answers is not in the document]
cavasinf: Twig don't allow to create multiple blocks with the same name.
As we discuss already, `if/else` twig condition, are not ignored when rendering template.

So it ended as a `The block 'navbar_start' has already been defined line x`.

This is the only work around I've found..

templates/layout-vertical.html.twig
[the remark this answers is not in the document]
kevinpapst: Hm, something must have changed, I am 100% sure that it was a problem in the past.
Good to know, I'll give it a try.

templates/layout-vertical.html.twig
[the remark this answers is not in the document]
kevinpapst: I have never seen blocks inside a macro, what is the difference when you remove them completely and only leave the content inside?

templates/layout-vertical.html.twig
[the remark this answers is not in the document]
cavasinf: Check with the Demo, you have override `navbar_start` and not `navbar_end`.
If what you said is true, you should have an error.

templates/layout-vertical.html.twig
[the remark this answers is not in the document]
cavasinf: > what is the difference when you remove them completely and only leave the content inside?

I have a slow brain during week-end, what do you mean by that, I don't understand.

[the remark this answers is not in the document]
kevinpapst: I mean: what happens when you have this code
```twig
{% macro navbar(navbar_start, navbar_notifications, navbar_user, navbar_end) %}
    {{ navbar_start ?? '' | raw }}
        {% if navbar_notifications is defined %}
            {{ navbar_notifications|raw }}
        {% else %}
            {% include '@Tabler/includes/navbar_notifications.html.twig' %}
        {% endif %}
        {% if navbar_user is defined %}
            {{ navbar_user|raw }}
        {% else %}
            {% include '@Tabler/includes/navbar_user.html.twig' %}
        {% endif %}
    {{ navbar_end ?? '' | raw }}
{% endmacro %}
```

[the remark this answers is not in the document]
cavasinf: > I have never seen blocks inside a macro

ATE, it is not really a `block`, only a string.
So the variables given to the macro are strings.
This is why `xxx | raw` filter is used, to convert string into pure HTML.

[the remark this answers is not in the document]
cavasinf: > what is the difference when you remove them completely and only leave the content inside?

~I have a slow brain during week-end, what do you mean by that, I don't understand.~

I get it now.

You mean if the macro would look like this.

```twig
{% macro navbar(navbar_start, navbar_notifications, navbar_user, navbar_end) %}
    {{ (navbar_start ?? '') | raw }}
        {% if navbar_notifications is defined %}
            {{ navbar_notifications|raw }}
        {% else %}
            {% include '@Tabler/includes/navbar_notifications.html.twig' %}
        {% endif %}
        {% if navbar_user is defined %}
            {{ navbar_user|raw }}
        {% else %}
            {% include '@Tabler/includes/navbar_user.html.twig' %}
        {% endif %}
    {{ (navbar_end ?? '') | raw }}
{% endmacro %}
```

It will be the same result.

Personally, I let block inside this kind of macro.
It allow me to have a friendly reminder that I can override blocks from extends.

We can remove it if you want.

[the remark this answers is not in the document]
tacman: I also like having the reminder there.

The includes can be done as a function, which is the recommended way, e.g.

```
 {{ include '@Tabler/includes/navbar_user.html.twig' }}
```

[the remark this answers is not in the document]
cavasinf: @tacman twig doesn't allow blocks override from extend, embedded from inside, include or use

kevinpapst:
@cavasinf I do not know, sorry 😄 too long ago and too many other projects in between, since my last review. 
I want to have it as well in my project and will look into it this week.

kevinpapst (dismissed):
Added some minor changes:
- remove global variable layout_type
- don't show badge on first level in vertical menu

Awesome stuff, thanks @cavasinf 👍 

I will release once I could test it in my app. Demo has no issues anymore!

kevinpapst (approved):

kevinpapst:
Wohoo 🎉 🚀

Issue #5, opened 2021-12-06
Login form - Tabulation navigation


cavasinf:
I can resolve that "issue" on my side by override some block,
but I want to know what is your (@kevinpapst) point of view about that.

In the Tabler demo : https://preview.tabler.io/sign-in.html
When you are in the input form, 
and try to move from `Email address` to `Password` by tabulation, 
it places you on `I forgot password`.

![LoginTabNavigation](https://user-images.githubusercontent.com/25293190/144813849-70434b5b-8ec8-4db4-97c2-e9f196f0a95d.gif)

It would be nicer if the tabulation action place us in the password input.

We can achieve that by adding `tabindex="-1"` in the HTML attr of the `<a>` tabler_password_reset link.

https://github.com/kevinpapst/TablerBundle/blob/5110ae631e47581880c668d456c20a9dd3d2bafa/templates/security.html.twig#L43

❓ The question is ❓ 
Do you allow in your bundle, extra functionalities/modifications that is not part from Tabler ?
Or do you try to keep as close as possible from the original ?

kevinpapst:
The tab order should be optimized for the most common case: Email, password, remember me, login, social links and forgot password as last option.
I'd say we should fix that upstream, the risk is that this change could be potentially lost in an upstream sync (unlikely but possible).
If upstream won't change it, we can do it here.

cavasinf:
Merged in https://github.com/kevinpapst/TablerBundle/pull/6

PR #6, opened 2021-12-06
Add `tabindex` on login form inputs

cavasinf:
## Description
Add `tabindex` on login form inputs

## Types of changes

## Checklist

### DIFF
--- a/templates/security.html.twig
+++ b/templates/security.html.twig
@@ -32,7 +32,7 @@
                     {% block login_form_start %}{% endblock %}
                         <div class="mb-3">
                             <label class="form-label">{{ 'Username'|trans({}, 'TablerBundle') }}</label>
-                            <input type="text" name="_username" class="form-control" placeholder="{{ 'Username'|trans({}, 'TablerBundle') }}" value="{{ last_username|default('') }}">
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thanks 😄  I will review/merge all PRs the next days when I have time to properly test it with my app. 
Please use your fork until then!

cavasinf:
#### Update 📋 
- Last required "input" (Sign In) is at index 40.
- Input `forgot password` is at index 100.

It allow devs to add 60 indexes between those inputs, is that enough ?

#### Question ❔ 
As _Originally posted by @kevinpapst in https://github.com/kevinpapst/TablerBundle/issues/5#issuecomment-986555655_ : **social links should have index too**.

I can put them at 50 and 60 by default. But that force devs to override the whole block if they want to change that.

As soon there are configurable/editable, is that correct to add a "default" index.
For example in my use case, I want Google auth and Facebook Auth. So I've removed the defaults one.

cavasinf:
@kevinpapst If it's OK for you, you can merge.

kevinpapst:
Ok, I tested it and believe that "Password forgotten" should be after the social logins. Why? Because if your app supports social logins, it is very likely that most users will use the OAuth approach and quickly want to reach those buttons and therefor never need the password forgotten link anyway.

kevinpapst (approved):
Nice approach 👍  
I am now using tab index 50 in my app for the oauth button (I only support for one provider).

PR #22, opened 2021-12-14
Add `before`/`after` on card blocks `header`/`footer`

cavasinf:
## Description
Add blocks `before` & `after` on `header/footer` (embeds/card) to allow devs to reproduce custom cards,
like `user card` on Tabler.io: https://preview.tabler.io/users.html

## Types of changes

## Checklist

### DIFF
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -15,6 +15,7 @@
     {% if _boxtype is not null %}
     <div class="card-status-top bg-{{ _boxtype }}"></div>
     {% endif %}
+    {% if block('box_header_before') is defined %}{{ block('box_header_before') }}{% endif %}
     {% if _title is not null or _collapsible or _tools is not null or _header is not null %}
     <div class="card-header">
         {% if _header is not null %}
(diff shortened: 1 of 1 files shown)


kevinpapst:
How would such a card look with the embed? I mean the twig code.
And the embed is already a monster 😁 adding 4 possible blocks is not going to make it any better^^

But my actual fear is that this will become a very tight coupling between embed and HTML. I could migrate > 90% of my app without any work, because I used this embed before in AdminLTE. But whenever I had some custom HTML in some embed block I needed manual adjustments. 

So maybe it would be better to create explicit embeds for special card (like the user)?

cavasinf:
For my case (code is in WIP), 
it's simple as the demo:

```twig
            {% embed '@Tabler/embeds/card.html.twig' with {boxtype :'primary'} %}
                {% block box_body %}
                    <div class="divide-y">
                        <div class="text-center">
                            <div class="avatar avatar-xl mb-3 avatar-rounded">
                                <i class="fa fa-building" style="font-size: 77px"></i>
                            </div>
                            <h3 class="m-0 mb-1"><a href="#">{{ tier.socialReason }}</a></h3>
                            <div class="text-muted">{{ tier.address }}</div>
                            <div class="mt-3">
                                <span class="badge bg-green-lt">tier.type</span>
                            </div>
                        </div>
                    </div>
                {% endblock %}
                {% block box_footer_before %}
                    <div class="d-flex">
                        <a href="#" class="card-btn">Email</a>
                        <a href="#" class="card-btn">Call</a>
                    </div>
                {% endblock %}
            {% endembed %}
```

image

kevinpapst (approved):

kevinpapst:
Yeah, why not. Let the dev decide how to use it 👍

PR #24, opened 2021-12-15
Card Embed - BugFix when using `_fullsize` and class name

cavasinf:
## Description
BugFix when using `_fullsize`.
Class `p-0` was attached to the previous class.

Same fix for `collapse` to be more safe.

## Types of changes

## Checklist


kevinpapst (approved):
I guess adjusting "collapse" was not necessary 😁 but `p-0` definitely was. Thanks!

PR #25, opened 2021-12-15
[Feature] Embed card - Allow footer to be collapsed

cavasinf:
## Description
Add the possibility to also collapse the footer on card embed.

### Changes 📋 
- New `embed` param: `footer_collapsible`
- Add private variable `_collapsible_class` to be more clear
- Change use of `_id` only for the body id
- Change `button.action_collapsebutton` target on class name instead of id

#### Openned
image

#### Before
image

#### After 
image

### How to use
```twig
{% embed '@Tabler/embeds/card.html.twig' with {collapsible : true, footer_collapsible : true} %}
    [...]
{% endembed %}
```

## Types of changes

## Checklist

### DIFF
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -13,7 +13,7 @@
 {% set _tools               = block('box_tools') is defined ? block('box_tools') : null %}
 {% set _header              = block('box_header') is defined ? block('box_header') : null %}
 
-<div class="card mb-3{% if block('box_class') is defined %} {{ block('box_class') }}{% endif %}"{% if block('box_attributes') is defined %} {{ block('box_attributes') }}{% endif %}>
+<div {% if _id is not null %}id="{{ _id }}"{% endif %} class="card mb-3{% if block('box_class') is defined %} {{ block('box_class') }}{% endif %}"{% if block('box_attributes') is defined %} {{ block('box_attributes') }}{% endif %}>
     {% if _boxtype is not null %}
     <div class="card-status-top bg-{{ _boxtype }}"></div>
     {% endif %}
@@ -38,7 +38,7 @@
     </div>
     {% endif %}
     {% if block('box_header_after') is defined %}{{ block('box_header_after') }}{% endif %}
-    <div {% if _id is not null %}id="{{ _id }}"{% endif %} class="{% if _collapsible %} {{ _collapsible_class }}{% if _collapsed %} collapse{% else %} show{% endif %}{% endif %} card-body{% if block('box_body_class') is defined %} {{ block('box_body_class') }}{% endif %}{% if _fullsize %} p-0{% endif %}">{{ block('box_body') }}</div>
+    <div class="{% if _collapsible %} {{ _collapsible_class }}{% if _collapsed %} collapse{% else %} show{% endif %}{% endif %} card-body{% if block('box_body_class') is defined %} {{ block('box_body_class') }}{% endif %}{% if _fullsize %} p-0{% endif %}">{{ block('box_body') }}</div>
     {% if block('box_footer_before') is defined %}{{ block('box_footer_before') }}{% endif %}
     {% if _footer and block('box_footer') is defined %}
         {# 
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -1,15 +1,17 @@
 {% import '@Tabler/components/buttons.html.twig' as button %}
 {% if block('box_before') is defined %}{{ block('box_before') }}{% endif %}
 
-{% set _collapsed   = collapsed is defined ? collapsed : false %}
-{% set _collapsible = collapsible is defined ? collapsible : _collapsed %}
-{% set _fullsize    = fullsize is defined ? fullsize : false %}
-{% set _footer      = use_footer|default(false) or block('box_footer') is defined %}
-{% set _boxtype     = boxtype|default(null) %}
-{% set _id          = id|default('collapse_' ~ random()) %}
-{% set _title       = block('box_title') is defined ? block('box_title') : null %}
-{% set _tools       = block('box_tools') is defined ? block('box_tools') : null %}
-{% set _header      = block('box_header') is defined ? block('box_header') : null %}
+{% set _collapsed           = collapsed is defined ? collapsed : false %}
+{% set _collapsible         = collapsible is defined ? collapsible : _collapsed %}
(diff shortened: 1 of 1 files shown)


cavasinf:
@kevinpapst I know you've said `card` was really a monster, but it's a must have feature.
And I think it's not too complex to implement it.

kevinpapst:
Hm, I never tried collapse on cards with footer. 
I wonder if we should instead make that default behavior. 
Is there a good use-case for only collapsing the body? 
I mean, there certainly is, but what is the 80:20 rule here? 
My gut feeling is that you want to collapse the entire card including the footer in 80% of the case, so we should make that default. 
Why did you remove the ID?

templates/embeds/card.html.twig
```diff
@@ -1,15 +1,17 @@
...
-{% set _id          = id|default('collapse_' ~ random()) %}
-{% set _title       = block('box_title') is defined ? block('box_title') : null %}
-{% set _tools       = block('box_tools') is defined ? block('box_tools') : null %}
-{% set _header      = block('box_header') is defined ? block('box_header') : null %}
+{% set _collapsed           = collapsed is defined ? collapsed : false %}
+{% set _collapsible         = collapsible is defined ? collapsible : _collapsed %}
+{% set _collapsible_class   = 'collapse_' ~ random() %}
+{% set _fullsize            = fullsize is defined ? fullsize : false %}
+{% set _footer              = use_footer|default(false) or block('box_footer') is defined %}
+{% set _footer_collapsible  = footer_collapsible is defined ? footer_collapsible : false %}
```
cavasinf, line 9: Changing `false` default value to `true` would collapse footer by default

templates/embeds/card.html.twig
```diff
@@ -28,15 +30,15 @@
...
-                        '#' ~ _id
+                        '.' ~ _collapsible_class
                     ) }}
                 {% endif %}
             </div>
         {% endif %}
     </div>
     {% endif %}
     {% if block('box_header_after') is defined %}{{ block('box_header_after') }}{% endif %}
-    <div {% if _collapsible %}id="{{ _id }}" {% endif %}class="{% if _collapsible %}{% if _collapsed %} collapse{% else %} show{% endif %}{% endif %} card-body{% if block('box_body_class') is defined %} {{ block('box_body_class') }}{% endif %}{% if _fullsize %} p-0{% endif %}">{{ block('box_body') }}</div>
```
cavasinf, line 39: It's only used for body id and only that.
Not used for collaspe event/action

templates/embeds/card.html.twig
```diff
@@ -1,15 +1,17 @@
...
-{% set _tools       = block('box_tools') is defined ? block('box_tools') : null %}
-{% set _header      = block('box_header') is defined ? block('box_header') : null %}
+{% set _collapsed           = collapsed is defined ? collapsed : false %}
+{% set _collapsible         = collapsible is defined ? collapsible : _collapsed %}
+{% set _collapsible_class   = 'collapse_' ~ random() %}
+{% set _fullsize            = fullsize is defined ? fullsize : false %}
+{% set _footer              = use_footer|default(false) or block('box_footer') is defined %}
+{% set _footer_collapsible  = footer_collapsible is defined ? footer_collapsible : false %}
+{% set _boxtype             = boxtype|default(null) %}
+{% set _id                  = id|default(null) %}
```
cavasinf, line 11: `id` still there

cavasinf (commented):
Question about `id`

templates/embeds/card.html.twig
cavasinf, line 9:
```suggestion
{% set _footer_collapsible  = footer_collapsible is defined ? footer_collapsible : true %}
```

cavasinf:
```suggestion
{% set _footer_collapsible  = footer_collapsible is defined ? footer_collapsible : true %}
```

templates/embeds/card.html.twig
cavasinf: @kevinpapst after using it, `id` is more logic to have on the `card` and not the `card-body`.
With the current state, targeting changes by the id, does not affect the footer.

templates/embeds/card.html.twig
kevinpapst: Yeah, but not filled. I know I have some automatic Javascript that fetches the ID and uses it for xyz (can't remember what it does 😁). 

It is simply a BC break that someone needs to be aware of. So it should be documented in the [migration guide](https://github.com/kevinpapst/TablerBundle/blob/main/docs/migration_guide.md) (I already added it). 

I have therefor added the first documentation files at https://github.com/kevinpapst/TablerBundle/tree/main/docs

templates/embeds/card.html.twig
kevinpapst: And what is your opinion on this? I'd say: in most cases you want to collapse the entire card

templates/embeds/card.html.twig
cavasinf: > Yeah, but not filled

I don't get it, how it is not filled ?
If you pass `id` parameter to the embed, works fine.

You maybe want a "default" id for your case ?
We can add that instead of `default(null)`, to like `default('card_' ~ random())`.
But I don't know if it's a good idea to auto implement id for card(s) embed.

templates/embeds/card.html.twig
kevinpapst: 👍

templates/embeds/card.html.twig
cavasinf: On AdminLTE it is the default behavior.
I don't mind it. Seems normal/logic to me.

templates/embeds/card.html.twig
kevinpapst: My JS relied on the ID and I did not pass it manually, because there is a potential unlimited number of cards. Thats why the auto ID was there in the first place. But it is okay, I can change my JS as well.

I don't even know if we should add the automatic class, now that you removed the automatic ID assignment. If we allow passing these values, it is the responsibility of the developer of the theme to take care of his requirements.

cavasinf: > it is the responsibility of the developer of the theme to take care of his requirements

I agree 👍

kevinpapst (approved):
I was blind with the id vs class issues. We need the class in two places, to collapse both divs.

LGTM, thank you @cavasinf 👍

PR #32, opened 2021-12-17
[Feature] Button macro - Simplified it

cavasinf:
## Description
I'm sorry but I have to 😢.

#### Changes:
- Simple way to disable icon from `button()` macro.
Give `false` to `icon` parameter.
- Add label color into macro `values` parameter.

### Demo
image

#### Twig
```twig
{{ button('save', { title : 'action.save'|trans}) }}
{{ button('save', { title : 'action.save'|trans}, false) }}
{{ button('save', { title : 'action.save'|trans, label: 'test'}) }}
{{ button('save', { title : 'action.save'|trans}, 'success') }}
{{ button(false , { title : 'action.save'|trans}) }}
{{ button(false , { title : 'action.save'|trans}, 'link') }}
{{ button(false , { title : 'action.save'|trans}, 'success') }}
{{ button('save', { title : 'action.save'|trans, combined: true}, 'success') }}
{{ button('save', { title : 'action.save'|trans, combined: true, label: 'test', labelColor: 'red'}, 'success') }}
```
#### HTML rendered
```html
<a class="btn  btn-icon" href="#" data-toggle="tooltip" data-bs-placement="top" title="Enregistrer"><i class="icon fas fa-save"></i></a>
<a class="btn  btn-icon" href="#" data-toggle="tooltip" data-bs-placement="top" title="Enregistrer"><i class="icon fas fa-save"></i></a>
<a class="btn" href="#" data-toggle="tooltip" data-bs-placement="top" title="Enregistrer"><i class="icon fas fa-save"></i><span class="badge bg-yellow ">test</span></a>
<a class="btn btn-success  btn-icon" href="#" data-toggle="tooltip" data-bs-placement="top" title="Enregistrer"><i class="icon fas fa-save"></i></a>
<a class="btn" href="#">Enregistrer</a>
<a class="btn btn-link" href="#">Enregistrer</a>
<a class="btn btn-success" href="#">Enregistrer</a>
<a class="btn btn-success" href="#"><i class="icon fas fa-save"></i>
    Enregistrer
</a>
<a class="btn btn-success" href="#"><i class="icon fas fa-save"></i>
    Enregistrer
    <span class="badge bg-red ms-1">test</span>
</a>
```

## Types of changes

## Checklist


templates/components/button.html.twig
```diff
@@ -22,9 +20,10 @@
         {% set attr = values.attr ?? {} %}
         {% set forceTitle = values.combined ?? false %}
         {% set label = values.label ?? null %}
```
cavasinf, line 22: I don't really like `label` as badge data.
I was really lost when I saw it (without doc), I though it was the label for the button.

IMO it needs to be called `badge` and `badgeColor`

templates/components/button.html.twig
```diff
@@ -1,8 +1,6 @@
 {% macro button(icon, values, type) %}
     {%- apply spaceless -%}
-        {% if type is same as (false) %}
-            {% set class = "" %}
-        {% elseif type is null %}
+        {% if type is null or type is same as (false) %}
```
cavasinf, line 3: I've let `type is same as (false)` for you (@kevinpapst) in case you still have condition on that.

Need to be removed on the next major version + deprecated DOC

cavasinf:
Still with some questions about the default operation with Icon VS Title display.

- IMO, Icon and Title should be "combined" by default.
if option `combined` is not given to macro, value must be `true` by default.
- Tabler prefer to use button without icon: https://preview.tabler.io/buttons.html.
- Today, button is only with icon and hover title.
Option `combined` has to be given to display title in button.

kevinpapst:
I have added a demo application (which is mostly broken by now) but it would be great if we could incorporate changes in there: https://github.com/kevinpapst/TablerBundle-Demo 

I added a page for the buttons and they look completely different then your screenshot:
Bildschirmfoto 2021-12-17 um 15 34 21

cavasinf:
@kevinpapst can I copy demo into my local ?
Looks like there are files missing like composer.json, webpack.config.js, ...

kevinpapst:
Lets talk about the demo here: https://github.com/kevinpapst/TablerBundle/discussions

kevinpapst:
After fixing stuff on my end I am here:
Bildschirmfoto 2021-12-17 um 17 35 13
Are you sure you committed everything?

cavasinf:
I've switch to `"kevinpapst/tabler-bundle": "dev-feature/button-embed"` from my git and it works fine on my side.

image

templates/components/button.html.twig
kevinpapst: may I ask why you changed that? this is on use-case less now

templates/components/button.html.twig
kevinpapst: i am using the entire macro in my application in hundreds of places. you can change it to
`{% set badge = values.label ?? (values.badge ?? null) %}` or even the other way around, but we need to keep the support for label

kevinpapst:
I have no idea what was wrong, but now I see your changes. Looks good to me, but please have a look at my comments before I merge

templates/components/button.html.twig
cavasinf: Yeah we can define it like that.
But in the other way, to recommend to devs to use badge instead.

`{% set badge = values.badge ?? (values.label ?? null) %}`

templates/components/button.html.twig
cavasinf: In what use case you define `type` as false ?

cavasinf:
@kevinpapst what do you think about that https://github.com/kevinpapst/TablerBundle/pull/32#issuecomment-996723633

cavasinf:
`badge` param has been added.

Example will looks like this:
```twig
    {{ button('save', { title : 'action.save'|trans, badge: 'test'}) }}
    {{ button('save', { title : 'action.save'|trans, combined: true, badge: 'test', badgeColor: 'red'}, 'success') }}
```

templates/components/button.html.twig
kevinpapst: The system is highly dynamic and used in combination with by an Event driven approach, which is again used by plugins (which I did not create) of my software. So I tried to find the places by triggering an exception if `type` is false, but I am in the middle of a refactoring and have too many other test failures right now... 
I need to fix these other errors first and then I can answer the question if I need support for false or not.
Maybe that was only needed in AdminLTE but is not needed in Tabler anymore, which I so believe, but can't prove.

kevinpapst (dismissed):
Look good, thanks 👍 
I will not directly merge, because I still want to find out if I need type=false. If not, we can remove it entirely.

kevinpapst:
I wanted to merge, but we changed the macro in another PR => can you merge main please?

cavasinf:
I've merged it. 👍

kevinpapst (approved):

kevinpapst:
Added some short docs for the macro. Thanks 🎉

PR #36, opened 2021-12-17
Allow Symfony 6

tacman:
## Description

Allow the bundle to be installed with a Symfony 6 application

## Types of changes

Since TablerBundle is new, I think we should drop support for 4.4 and only support 5.4+ 
This simply allows the bundle to be installed with Symfony 6.

## Checklist
- [X] I agree that this code will be published under the [MIT license](https://github.com/kevinpapst/TablerBundle/blob/master/LICENSE)


kevinpapst:
Something is really off in this PR. Please rebase on latest master.

Issue #38, opened 2021-12-18
Menu badges not rendering


tacman:
```php
// MenuBuilderSubscriber.php
        $forms = new MenuItemModel('forms', 'Forms', null, [], 'fab fa-wpforms');
        $forms->setBadge('1');
        $forms->setBadgeColor('red');
```

but the red badge isn't rendering

![image](https://user-images.githubusercontent.com/619585/146644349-45466cb5-d713-422a-ad14-b5bba9332cfe.png)

kevinpapst:
Fixed by #40, thanks for the report!

PR #39, opened 2021-12-18
Render badge in menu

tacman:
## Description

Renders the menu item badge and badge class, first step in fixing #38 

I deprecated badge color and replaced it with badge class.

While the badge now renders, there's a conflict with the .badge css in tabler.css, and I'm not sure how to fix it.  So it's functional but not very pretty.

## Types of changes
- [ X ] Bug fix (non-breaking change which fixes an issue)
- [ X ] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/src/Model/MenuItemModel.php
+++ b/src/Model/MenuItemModel.php
@@ -49,8 +49,31 @@ class MenuItemModel implements MenuItemInterface
     private $badge;
     /**
      * @var string|null
+     * @deprecated
      */
     private $badgeColor;
+
+    /**
+     * @var string|null
+     */
+    private $badgeClass;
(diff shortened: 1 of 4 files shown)


kevinpapst:
Thank you for the idea 👍 

Looks a bit crappy 😁 not your fault obviously(!), but it seems the menu is not made for badges with content inside. 

Bildschirmfoto 2021-12-18 um 22 31 46

I would prefer this version, but it is up to the developer if the badge should contain some content or not.
This is simply a badge with only a color applied.

Bildschirmfoto 2021-12-18 um 22 35 56

See #40

Issue #21, opened 2021-12-14
Breadcrumb that is not a real breadcrumb


cavasinf:
## Context 📋 
In the `page-header`,
I want to add buttons like `Create new`, 
![image](https://user-images.githubusercontent.com/25293190/145976473-695e6134-692c-40e6-b28a-f319b096a6a2.png)

and still have the breadcrumb: `Items > Actions`.
![image](https://user-images.githubusercontent.com/25293190/145976527-76df4e0a-762d-4d4f-9511-48e1c45f97ec.png)

## Current state 👀 
Page header buttons are at the position of `Breadcrumb`.
https://github.com/kevinpapst/TablerBundle/blob/014debb4a80785a99d070259b62cc34d4be2ad90/templates/layout.html.twig#L96-L109

## Problem ❌ 

With that current state,
- If I want to add my buttons, I'll put them INSIDE breadcrumb block with custom HTML to allow to have Buttons + Breadcrumb.
- Plus, we are declaring a block about Breadcrumb with buttons on default values. In that case, it's not really a "Breadcrumb".

## Possible solution ✔️ 

Split current `{% block breadcrumb %}` into 3 blocks
```twig
{% block page_header_right %}
    <div class="col-auto ms-auto d-print-none">
        {% block page_header_actions %}
            <div class="btn-list">
                <a href="#" class="btn btn-white">
                    New view
                </a>
                <a href="#" class="btn btn-primary d-none d-sm-inline-block">
                    Create new report
                </a>
            </div>
        {% endblock %}
        {% block breadcrumb_container %}
            <div class="text-end">
                {% block breadcrumb %}
                    Entity > Action > ...
                {% endblock %}
            </div>
        {% endblock %}
    </div>
{% endblock %}
```

cavasinf:
Other solution could be to use `page_stats` as breadcrumb.
And rename `breadcrumb` into `page_header_actions`

![image](https://user-images.githubusercontent.com/25293190/145990640-29a7480d-f150-44a6-af8b-a4407f76f23b.png)

kevinpapst:
Hm, I would always try to keep the menu level as flat as possible. 
And then you could switch page title and pre title and add a link to the pre title - in your example: `Articles` would be the `pre_title` with a link and `Create article` would be the title.

The other solution I would prefer are be the 3 blocks. 
But I would keep the breadcrumb on the left side (start for RTL) and followed by the buttons on right side (end for RTL).

[...]

Issue #19, opened 2021-12-10
Title block names - Wrongly named


cavasinf:
When using the blocks for header titles, there names can be not comprehended by devs (and me).

We are talking about `subtitle` and `title`.
Or in Tabler, it's more `pretitle` and `title`.

### Example 
My page is a `List of products` that come from `Products`.
Breadcrumb would be : `Products > List of products`.

|  | Tabler.io | TablerBundle |
|:---:|:---:|:---:|
| Variable | ![image](https://user-images.githubusercontent.com/25293190/145596355-f35551ea-1e04-459f-87fa-3928bebd9b7b.png) | ![image](https://user-images.githubusercontent.com/25293190/145595699-7aa35aa8-f060-4f89-bc9e-9f5fc94169ff.png) |  
| Example | ![image](https://user-images.githubusercontent.com/25293190/145596646-d60b075f-f734-441a-a9e3-0f449cd57a69.png) | ![image](https://user-images.githubusercontent.com/25293190/145596764-3b037e63-f2f3-40fc-8ef5-1d3abc44fbb2.png) |

I think it's more normal/user friendly to have the main title more bigger and wider than the pretitle.

⚠️ Renaming a block can break existing projects ⚠️

cavasinf:
@kevinpapst any thought on that ?

kevinpapst:
There is no BC policy that we yet have to stick to. This bundle is WIP.
I just don't understand the issue. You want to switch block names?

cavasinf:
Yes, I think it's better to stick with Tabler variable name:
Switch `page_subtitle` into `page_pretitle`.

## Explication

I guess, the current names are a rest from AdminLTE. 
Which works for AdminLTE template:
![image](https://user-images.githubusercontent.com/25293190/145805450-1b0f93ca-b0e8-4f1e-8f84-c0be6b9fe04e.png)

But if we stick with those today variable names, 
it look like kinda reversed between title and subtitle.
![image](https://user-images.githubusercontent.com/25293190/145805674-0e278696-c4c0-424c-85f8-29214d0070e9.png)

kevinpapst:
Yes, the names make the transition easier and I don't think it really matters.
That is a bit like the discussion about CSS class names: name it after "what it does" or "where it is used"?

Using `pre` explains where it is placed and `sub` describes that this is an additional info about the current page.

Personally I'd always say I am on the page "General UI" here. Do you want to make "Prevew of UI Elements" big in this example?

[...]

PR #48, opened 2022-01-04
[BugFix] Reorder blocks to match breadcrumb from Tabler.io doc

cavasinf:
## Description
According to #19 & #21

Renaming blocks for breadcrumb.

From:
image
To:
image

---------------

From:
image
To:
image

## Types of changes

## Checklist

### DIFF
--- a/templates/layout.html.twig
+++ b/templates/layout.html.twig
@@ -82,9 +82,11 @@ Enjoy your theme!
             <div class="page-header d-print-none">
                 <div class="row align-items-center">
                     <div class="col">
+                        {% block breadcrumb %}
                         <div class="page-pretitle">
-                            {% block page_pretitle %}{% block breadcrumb %}Overview{% endblock %}{% endblock %}
+                            {% block page_pretitle %}Overview{% endblock %}
                         </div>
+                        {% endblock %}
                         <h2 class="page-title">
                             {% block page_title %}Dashboard{% endblock %}
                         </h2>
--- a/templates/layout.html.twig
+++ b/templates/layout.html.twig
@@ -83,19 +83,19 @@ Enjoy your theme!
                 <div class="row align-items-center">
                     <div class="col">
                         <div class="page-pretitle">
-                            {% block page_subtitle %}Overview{% endblock %}
+                            {% block page_pretitle %}{% block breadcrumb %}Overview{% endblock %}{% endblock %}
                         </div>
                         <h2 class="page-title">
                             {% block page_title %}Dashboard{% endblock %}
                         </h2>
-                        {% block page_stats %}
+                        {% block page_stats %}{% block page_subtitle %}
                         <div class="text-muted mt-1">
                             1-10 of 100
                         </div>
-                        {% endblock %}
+                        {% endblock %}{% endblock %}
                     </div>
                     <div class="col-auto ms-auto d-print-none">
-                        {% block breadcrumb %}
+                        {% block page_header_actions %}
                         <div class="btn-list">
                             <a href="#" class="btn btn-white">
                               New view
--- a/templates/layout.html.twig
+++ b/templates/layout.html.twig
@@ -90,7 +90,7 @@ Enjoy your theme!
                         <h2 class="page-title">
                             {% block page_title %}Dashboard{% endblock %}
                         </h2>
-                        {% block page_stats %}{% block page_subtitle %}
+                        {% block page_subtitle %}
                         <div class="text-muted mt-1">
                             1-10 of 100
                         </div>
--- a/templates/layout.html.twig
+++ b/templates/layout.html.twig
@@ -94,7 +94,7 @@ Enjoy your theme!
                         <div class="text-muted mt-1">
                             1-10 of 100
                         </div>
-                        {% endblock %}{% endblock %}
+                        {% endblock %}
                     </div>
                     <div class="col-auto ms-auto d-print-none">
                         {% block page_actions %}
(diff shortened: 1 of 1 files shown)


templates/layout.html.twig
```diff
@@ -83,19 +83,19 @@ Enjoy your theme!
...
                     <div class="col">
                         <div class="page-pretitle">
-                            {% block page_subtitle %}Overview{% endblock %}
+                            {% block page_pretitle %}{% block breadcrumb %}Overview{% endblock %}{% endblock %}
                         </div>
                         <h2 class="page-title">
                             {% block page_title %}Dashboard{% endblock %}
                         </h2>
-                        {% block page_stats %}
+                        {% block page_stats %}{% block page_subtitle %}
```
cavasinf, line 91: At the end, `page_stats` is no longer needed. `page_subtitle` is enough.

@kevinpapst Do you want to keep it like that or do we remove `page_stats` ?

cavasinf (commented):
Need review

templates/layout.html.twig
kevinpapst: Yeah, if we do a BC break, then fully: you can remove the block completely.

templates/layout.html.twig
```diff
@@ -82,20 +82,22 @@ Enjoy your theme!
...
                         <div class="text-muted mt-1">
                             1-10 of 100
                         </div>
-                        {% endblock %}
+                        {% endblock %}{% endblock %}
                     </div>
                     <div class="col-auto ms-auto d-print-none">
-                        {% block breadcrumb %}
+                        {% block page_header_actions %}
```
kevinpapst, line 100: The other two are called `header_X`, then lets call this one `header_actions` as well, to be consistent

cavasinf: IDK if it's a good idea, It's inside of `{% block page_header %}`.

With PR https://github.com/kevinpapst/TablerBundle/pull/3:
- `<header>` has block `header_id`.
- `.page-header` has block `page_header_id`.

At the end, `header_actions` sounds like action in the menu.
`page_header_X` relate to the header of the current page.

kevinpapst: Then rename the other ones?

cavasinf: That is the question 😉 
> The other two are called header_X

Who are the two ? In `main` branch or PR ?

kevinpapst: From what I can see the block names are: `page_pretitle`, `page_stats`, `page_title`, `page_subtitle`, `page_header_actions` and `breadcrumb`. 
Bildschirmfoto 2022-01-04 um 14 10 43

We can most certainly find a consistent naming, can't we? Now that we change it once and for all.

cavasinf: I see now,
We can call them `page_X` when it depend on a specific page. 👍 

At the end we can:
- Leave like that: `page_pretitle`, `page_stats`, `page_title`, `page_subtitle`.
- Remove `page_stats`.
- And change `page_header_actions` to `page_actions`.

But now you make me question about `breadcrumb` 😄
Should we call him `page_breadcrumb` ? It will be a BC Break for you..
`breadcrumb` simple is fine to me.

kevinpapst: I don't care about the BC breaks on my end. It is better to make it correct ones and then never have to touch it again. 

Hm, `breadcrumb` sounds correct to me as well, but with the same argument we could call it `actions` as well.
The other blocks are called `page_content_before `, `page_content` ... so we should either remove `page_` from all of them (brutal BC break, please not) or prefix the breadcrumb as well to `page_breadcrumb`. 

Honestly: I haven't thought about it before, but now that we discuss I'd like to have it consistent.

cavasinf: `page_content_before` , `page_content`, ... is not a problem to me.
Once again, there are targeting `page` configuration.
Like `page_wrapper_id` and `.page-body`. 
It's fine to me.

To conclude, I would leave `breadcrumb` as it, and let `page_actions` too.
But if you want to rename `page_actions` into `actions`, it is all yours.

cavasinf, line 100:
```suggestion
                        {% block page_actions %}
```

cavasinf:
```suggestion
                        {% block page_actions %}
```

templates/layout.html.twig
```diff
@@ -82,20 +82,22 @@ Enjoy your theme!
...
                         <div class="page-pretitle">
-                            {% block page_subtitle %}Overview{% endblock %}
+                            {% block page_pretitle %}Overview{% endblock %}
                         </div>
+                        {% endblock %}
                         <h2 class="page-title">
                             {% block page_title %}Dashboard{% endblock %}
                         </h2>
-                        {% block page_stats %}
+                        {% block page_stats %}{% block page_subtitle %}
```
kevinpapst, line 93: Lets remove page_stats then

cavasinf, line 93:
```suggestion
                        {% block page_subtitle %}
```

templates/layout.html.twig
```diff
@@ -82,20 +82,22 @@ Enjoy your theme!
...
                         <div class="text-muted mt-1">
                             1-10 of 100
                         </div>
-                        {% endblock %}
+                        {% endblock %}{% endblock %}
```
cavasinf, line 97:
```suggestion
                        {% endblock %}
```

cavasinf (commented):
Removing page_stats

kevinpapst:
Clever idea to use the suggestions, sorry totally forgot, oerthwise I would have done it myself 😁

PR #52, opened 2022-01-10
Fixes (best PR name in the world!)

### LINKED ISSUE #49: Card always have padding bottom
https://github.com/kevinpapst/TablerBundle/blob/3fc999c31d117b68bfed5860642f74d138193122/templates/embeds/card.html.twig#L16

With the current state, there's no way to remove padding bottom from embeded card.

```html
 <div ... class="card mb-3" ... >
```

How it looks today:
image

What I want to achieve:
image

kevinpapst:
## Description

- Fixes #49 
- Fix homepage link in error template 
- Added variable to customize card bottom margin

## Types of changes

## Checklist
- [x] I updated the demo app

### DIFF
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -12,9 +12,9 @@
 {% set _title               = block('box_title') is defined ? block('box_title') : null %}
 {% set _tools               = block('box_tools') is defined ? block('box_tools') : null %}
 {% set _header              = block('box_header') is defined ? block('box_header') : null %}
-{% set _margin_bottom       = margin_bottom is defined ? margin_bottom : 'mb-3' %}
+{% set _margin_bottom       = margin_bottom is defined ? margin_bottom : 3 %}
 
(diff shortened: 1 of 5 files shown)


templates/embeds/card.html.twig
```diff
@@ -12,8 +12,9 @@
 {% set _title               = block('box_title') is defined ? block('box_title') : null %}
 {% set _tools               = block('box_tools') is defined ? block('box_tools') : null %}
 {% set _header              = block('box_header') is defined ? block('box_header') : null %}
+{% set _margin_bottom       = margin_bottom is defined ? margin_bottom : 'mb-3' %}
```
cavasinf, line 15: Something tickle me with the variable name VS what's in.

With the current code, `margin_bottom` sounds like true/false. 
But it expect class name.

Without seeing the code, I was expecting a number to be given.
Like `{margin_bottom : 3}`.

What do you think ?
```twig
{% set _margin_bottom       = margin_bottom is defined ? 'mb-' ~ margin_bottom : 'mb-3' %}
```

kevinpapst: I am fine with your solution: if there is a `mb-0` class!? 
I just made it this way, so it can be removed entirely by providing an empty string (which is not very elegant tbh).

cavasinf: There is a mb-0 class.

cavasinf (commented):
Looks good to me!

PR #55, opened 2022-01-11
Add theme override example/doc

cavasinf:
## Description
Add theme override example/doc

## Types of changes

## Checklist


kevinpapst (changes requested):
That is a good starting point, but not yet enough for merging.
We should show at which URI the file must live and what else is necessary (packages.json, webpack, yarn/npm ...).
I have some half baked docs here which I will merge with yours.

kevinpapst (approved):
More docs are always better, one step at a time 😁

kevinpapst:
Thanks @cavasinf 👍

PR #56, opened 2022-01-13
[BugFix] Update flash to Tabler template

cavasinf:
## Description
Update flash template. There were not displayed correctly.

https://preview.tabler.io/docs/alerts.html#dismissible-alerts

Before (not visible):
image

After:
image
image

## Types of changes

## Checklist


kevinpapst (approved):
Awesome 👍 I probably would have never found that, because I overwrite that template with my own code using Sweetalert.
I would (for the future) propose a new twig filter to map `type` to `class`. Because in the past all my error flashes had type "error"  now it needs to be "danger", which I find weird.

PR #60, opened 2022-01-24
use tabler JS compatible tooltip, allow to overwrite the attribute

### LINKED ISSUE #58: [Button] Bootstrap tooltip
https://github.com/kevinpapst/TablerBundle/blob/8af12de25e69acc6658fed69eb331a2c532b6f47/templates/components/button.html.twig#L43

If using Tabler JS import file: 
- `data-toggle="tooltip"` supposed to be `data-bs-toggle="tooltip"`
https://github.com/tabler/tabler/blob/74985f8d008395d695d5ab1ad5b729684b00b8af/src/js/src/tooltip.js#L3

Should we use Tabler default behavior or Bootstrap configuration ?

kevinpapst:
## Description

Fixes #58

## Types of changes

## Checklist


cavasinf (commented):
LGTM 👍

PR #69, opened 2022-03-01
change argument order to simplify icon usage outside buttons

### LINKED ISSUE #68: `tabler_icon` change $default param
Hey @kevinpapst,

wondering what is the purpose of `$default` in the `tabler_icon` TWIG function.
https://github.com/kevinpapst/TablerBundle/blob/8a70eaa6d8d41b116803d0fe43a7f184210324c1/src/Twig/RuntimeExtension.php#L131-L134

Due to that, 
all of my calls of this function (when not using buttons), are made with `null` for that parameter.
```twig
{{ tabler_icon('email', null, false) }}
```

Two questions arise from this:
1. Should we use `false` to remove `.icon` class.
2. Should we override `.icon` css when using it outside of `.btn`.

My current problem is that the render with `icon` class and font-awesome is kinda ugly:
- Too big
- Not vertically centered
- Overlap with label when large icon (e.g : `fas fa-cogs`)

Today `.icon` class: 🥉 
image

`.icon` class overridden: 🥈 
image

Without `.icon` class: 🥇 
image

WDYT ?

_Originally posted by @cavasinf in https://github.com/kevinpapst/TablerBundle/discussions/67_

kevinpapst:
## Description

Attention: This can a BC break if you used more than an argument in the `|icon` filter or more than one argument in the `tabler_icon()` function before.

Fixes #68  

## Types of changes

## Checklist

### DIFF
--- a/templates/components/buttons.html.twig
+++ b/templates/components/buttons.html.twig
@@ -7,20 +7,12 @@
     {{ _self.action_cardtoolbutton('collapse', {collapse: target, title: label|default('')}) }}
 {% endmacro %}
 
-{# TODO test me #}
-{% macro link_toolbutton(icon, href, label, attrs) %}
-    {% set attrs = attrs|default({})|merge({class: 'me-2 ' ~ (attrs.class|default(''))}) %}
-    <a href="{{ href }}" {% for name, value in attrs|default({}) %} {{ name }}="{{ value }}"{% endfor %}>
-        {{ tabler_icon(icon) }}{% if label is not null %} {{ label }}{% endif %}
-    </a>
-{% endmacro %}
-
 {% macro link_button(label, href, icon, type, size) %}
     {% set _size = size|default(null) %}
     {% set _type = type|default('primary') %}
     <a href="{{ href|default('#') }}" class="btn btn-{{ _type }}{% if _size %} btn-{{ _size }}{% endif %}">
         {% if icon %}
-            {{ tabler_icon(icon) }}
+            {{ tabler_icon(icon, true) }}
         {% endif %}
         {% if label %}
             <span>{{ label }}</span>
@@ -33,7 +25,7 @@
     {% set _type = type|default('primary') %}
     <button data-action="{{ action }}" class="btn btn-{{ _type }}{% if _size %} btn-{{ _size }}{% endif %}">
         {% if icon %}
-            {{ tabler_icon(icon) }}
+            {{ tabler_icon(icon, true) }}
         {% endif %}
         {% if label %}
             <span>{{ label }}</span>
--- a/src/Twig/RuntimeExtension.php
+++ b/src/Twig/RuntimeExtension.php
@@ -123,13 +123,13 @@ final class RuntimeExtension implements RuntimeExtensionInterface
         return $userEvent;
     }
 
-    public function createIcon(string $name, bool $isButton = true, string $default = null): string
+    public function createIcon(string $name, bool $withIconClass = false, string $default = null): string
     {
-        return '<i class="' . $this->icon($name, $isButton, $default) . '"></i>';
+        return '<i class="' . $this->icon($name, $withIconClass, $default) . '"></i>';
     }
 
-    public function icon(string $name, bool $isButton = true, string $default = null): string
+    public function icon(string $name, bool $withIconClass = false, string $default = null): string
     {
-        return ($isButton ? 'icon ' : '') . ($this->icons[str_replace('-', '_', $name)] ?? ($default ?? $name));
+        return ($withIconClass ? 'icon ' : '') . ($this->icons[str_replace('-', '_', $name)] ?? ($default ?? $name));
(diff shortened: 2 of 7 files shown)


cavasinf:
> So having the icon class by default made sense to me.

How do you render your buttons in your app ? from `button()` twig macro or `buttons.html.twig` ?

If it's the case, we should rename `$isIcon` to `$hasIconClass` at `false` by default.
And set this param at `true` in the button macro.

I know it doesn't make sense for you but :
- `.icon` class in Tabler.io was made for `<svg>` not `<i>`.
- `.nav-link-icon` is used for menu.
- `.icon-lg` and other seems to not working (due to `width` in tabler and `font-size` in FontAwesome).
- didn't tried with `.card-stamp`.

Looks like I'm against you, but I'm not, just exposing my opinion and trying to have yours 😉

kevinpapst:
> Looks like I'm against you

Totally not! Its valuable feedback, but why did you say "I agree" in the discussion 😁 

I am rendering most buttons/actions with the macros, but some just don't work easily due to custom markup (data-attributes and such). 

I just changed the implementation in my app and had to adjust maybe 15 icon calls. Most of them were not changed because it was necessary, but just because they looked even better without the `icon` class (which makes them bigger, what I didn't realize at the time).

So you are arguing that users of the framework should use the button macro and then get the correct classes.
When the user calls `icon` manually, then we suppose it is not a button => meaning we switch `$isButton` to false by default?

cavasinf:
> but why did you say "I agree" in the discussion

Yeah sorry, I've always a part of my brain that works in background without the possibility of controlling it.
(hard to sleep because of that). And it denied what I've think/analyzed before 😅 

> but some just don't work easily due to custom markup

On my side, your macro fill every demands on my side. Even for SWUP or Stimulus and others.
If you want we can talk about your "problems".

> So you are arguing that users of the framework should use the button macro and then get the correct classes. And when the user calls icon manually, then we suppose it is not a button => meaning we switch $isButton to false by default?

Exactly 👍 Can't summarize more efficiently than that. 
The question is, do we rename `$isButton` 🤔

cavasinf:
OK so, my "multi-task" brain talked to me again.

I was looking at [Tabler.io CSS](https://github.com/tabler/tabler/blob/main/dist/css/tabler.css) for every cursors with `.icon` but targeting with a parent (like `.btn .icon`) and where there's not only a `width` / `height` css changes.

There's all cases:
- `.btn`
- `.btn-link`
- `.btn-icon`
- `.btn-action`
- `.card-stamp-icon` ❌ 
- `.form-selectgroup-label`
- `.list-timeline-icon`
- `.list-group-transparent .list-group-item`
- `.ribbon`
- `.status`
- `.chart-sparkline-label`
- `.badge`

ATE, it is not only for buttons 😭

kevinpapst:
Switch to tabler icons? Might be a bit more work though...

cavasinf:
> Switch to tabler icons? Might be a bit more work though...

No, this is not what I am thinking of..
I'm using FontAwesome since 7 years now, and it fill every needs that I have even with the free version.
I don't really like from Tabler, the fact they force us to use THERE icons.

I don't know how if it's a great idea, but we can "maybe" create our own CSS to handle FontAwesome in that specific case.
But that mean we are creating a "technology watch" of tabler and impact our code at every tabler update.

Second option is to warn Tabler of that issue, and ask them to allow FontAwesome.
But with all responses that I received from them, I don't think they will take care of that..
"Our problem not theirs"

kevinpapst:
timeline looks good with and without, what do you say?
the "trash bin" has `icon` class the first and last one doesn't:
Bildschirmfoto 2022-03-02 um 13 16 25

kevinpapst:
Thanks for the detailed investigation btw 👍 

I updated the PR. The demo is working again with some small tweaks.

Now I will check my app and replace all manual links with the macros: should fix/remove 90% of my current icon calls.

Now it is your turn: please review and test 😄

templates/components/buttons.html.twig
```diff
@@ -7,20 +7,12 @@
     {{ _self.action_cardtoolbutton('collapse', {collapse: target, title: label|default('')}) }}
 {% endmacro %}
 
-{# TODO test me #}
```
kevinpapst, line 10: wasn't used anywhere, so I removed it for now

cavasinf:
The most problematic icons are those who are not a "square".
Like:
- fas fa-users-cog
- fas fa-cogs
- fas fa-users

If those works, others too 👍

templates/components/buttons.html.twig
cavasinf: 👍

cavasinf (approved):
LGTM, nice catch with the self re-use macro for link and normal button 🥇

kevinpapst:
It was time for a new tag: https://github.com/kevinpapst/TablerBundle/releases/tag/0.2

PR #74, opened 2022-03-26
Change .wrapper to .page

cavasinf:
## Description
In case of the footer that is not on the bottom of the page, 
the only diff I've found with the demo is that class: https://preview.tabler.io/empty.html

Maybe BC Break ? not on my side.

Note: in the `fullpage.html.twig`, we have already changed that class.

With `.wrapper`:
Capture d’écran 2022-03-26 à 15 11 58

With `.page`:
Capture d’écran 2022-03-26 à 15 11 49

## Types of changes

## Checklist


kevinpapst (approved):
Nice catch!

PR #77, opened 2022-05-11
Update layout.html.twig

wucherpfennig:
Fix actions display on small screens.

grafik

### DIFF
--- a/templates/layout.html.twig
+++ b/templates/layout.html.twig
@@ -96,7 +96,7 @@ Enjoy your theme!
                                 </div>
                             {% endblock %}
                         </div>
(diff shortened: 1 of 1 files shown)


kevinpapst:
Thanks! Is that before or after? What has changed?

wucherpfennig:
Sorry I was lazy but in my defense it was late 😉

Actually this is the "new" behaviour. 

While updating AdminLTE to Tabler i noticed that the actions do not go below the header once reached the mobile viewport.

You can see the "right" behaviour here: https://preview.tabler.io

kevinpapst (approved):

PR #82, opened 2022-05-24
Add `status indicator` component

cavasinf:
See https://preview.tabler.io/docs/statuses.html#status-indicator
#### Usage

```twig
{% from '@Tabler/components/status_indicator.html.twig' import status_indicator %}

<div class="row align-items-center text-start">
    <div class="col p-0">
        {{ status_indicator({color : item.open ? 'green' : 'red', animated : false, extraClass :'float-end' }) }}
    </div>
    <div class="col p-0">
        {{ item.open ? 'label.opened'|trans : 'label.closed'|trans }}
    </div>
</div>
```

#### Preview
image

## Types of changes

## Checklist

### DIFF
--- a/templates/components/status_indicator.html.twig
+++ b/templates/components/status_indicator.html.twig
@@ -1,0 +1,13 @@
+{% macro status_indicator(options) %}
+
+    {% set _color = options.color ?? 'green' %}
+    {% set _animated = options.animated ?? true %}
+    {% set _extraClass = options.extraClass ?? '' %}
+
+    <span class="status-indicator {{ _animated ? 'status-indicator-animated' : '' }} {{ 'status-'~_color }} {{ _extraClass }}">
+      <span class="status-indicator-circle"></span>
+      <span class="status-indicator-circle"></span>
+      <span class="status-indicator-circle"></span>
+    </span>
+
+{% endmacro %}


kevinpapst:
That is a good question, which I cannot answer without research. 
I don't like the idea of adding them to the repo ... we would have to ignore the directory at least via .gitattributes to prevent the composer package from growing in size.
Maybe it is easiest, if we include them as visible image in the PR that adds the docs page - I hope visible images will not be deleted by some automated processes.

cavasinf:
Looking at my other PR for timeline Doc #83 
The Github static link show MY user ID. So we know who uploaded it.

When looking at the [Symfony repo for example](https://github.com/symfony/symfony), their logo is not under `user-images.githubusercontent.com` but `camo.githubusercontent.com`. 
User is not showed in that link.

Looking at gihub doc, there is an explanation for camo subdomain : https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls

kevinpapst:
> The Github static link show MY user ID. So we know who uploaded it.

Do you care? 😁 

I just want to assure that the image does not vanish because some process recognizes that it is not used anywhere.
But I don't care how we do it. Having it in a comment seems legit to me.

kevinpapst (approved):

cavasinf:
> Do you care?

No I don't 😉 

> Having it in a comment seems legit to me

What do you mean by that ?

kevinpapst:
> What do you mean by that ?

Not really sure hahaha  

I meant: as you did it in this PR. Upload the image how ever you like, make it part of the PR comment (like the status indictor image above) and then use it.

Probably to much talk about images^^ what do you think about a new release?

kevinpapst:
Yes, we can release as often as we want. For now I will mark them as "Pre-Release" as in "unstable before 1.0".

But if you plan too add more stuff this week, then I will wait a couple of days.

cavasinf:
> But if you plan too add more stuff this week, then I will wait a couple of days.

I'm in holiday this Wednesday (for most of french people), 
so go ahead and do what you plan to do. I'll follow.

PR #83, opened 2022-05-24
Add preview screenshot

cavasinf:
## Description
Add doc example, because I've forgot what does what.
So more doc is welcome 😉 

image | image

## Types of changes

## Checklist


kevinpapst:
Hahaha 😁
Btw. the demo app includes timeline examples, so you could have just looked it up there...
Where is this image located? Was it an comment upload? Do you know if they exist forever?

PR #84, opened 2022-05-24
Add steps component

cavasinf:
## Description
See https://preview.tabler.io/docs/steps.html

#### Usage
```twig
{% from '@Tabler/components/steps.html.twig' import steps %}

{{ steps([
    {title : 'Step 1', clickable : false},
    {title : 'Step 2'},
    {title : 'Step 3', active : true},
    {title : 'Step 4', description : 'More detailed description'},
    {title : 'Step 5', href: 'https://www.google.com', clickable : true},
],{
    tooltip : true,
    color: 'red',
    counter : true,
}) }}
```

#### Preview

steps2

## Types of changes

## Checklist

### DIFF
--- a/templates/components/steps.html.twig
+++ b/templates/components/steps.html.twig
@@ -1,0 +1,35 @@
+{% macro steps(steps, options) %}
+
+    {% set _tooltip = options.tooltip ?? false %}
+    {% set _color = options.color ?? 'primary' %}
+    {% set _counter = options.counter ?? false %}
+    {% set _attr = options.attr ?? '' %}
+
+    <div class="steps {{ _counter ? 'steps-counter' : '' }} {{ 'steps-'~_color }}" {{ _attr|raw }}>
+
+        {% set activeStepProcessed = false %}
+
+        {% for step in steps %}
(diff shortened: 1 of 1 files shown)


templates/components/steps.html.twig
```diff
@@ -0,0 +1,35 @@
...
+
+    <div class="steps {{ _counter ? 'steps-counter' : '' }} {{ 'steps-'~_color }}" {{ _attr|raw }}>
+
+        {% set activeStepProcessed = false %}
+
+        {% for step in steps %}
+            {% set _type = (step.clickable ?? not activeStepProcessed) ? 'a' : 'span' %}
+            {% set _title = step.title ?? '' %}
+            {% set _description = step.description ?? _title %}
+            {% set _active = step.active ?? false %}
```
kevinpapst, line 16: Does that work if someone uses `active: false`?

kevinpapst: Forget it, I need coffee...

kevinpapst (approved):
Awesome component. I can use that immediately.
We should probably start adding the URLs to the tabler docs in the component file.

cavasinf:
You are too fast 😆

Issue #85, opened 2022-05-24
Macro utilities ?


cavasinf:
Due to #84 

I was searching in the Bundle, what is the best way to handle custom `attr` for components.

I really like the `button` component way.
https://github.com/kevinpapst/TablerBundle/blob/ea98cb05ce72cc3314ecc1572126e4bc5bdd7af1/templates/components/button.html.twig#L49-L53

But the code is "heavy" to copy pasta between component and child component.

I was implementing a macro for this component only, 
but it will be nice if we do utilities that can be called between all components.
Twig `Filter` or `Function` is maybe overpower.

Maybe a simple twig file under `/templates/components/macros.html.twig` ❔ 

Example macro: 
```twig

{% macro attr_to_html(attr) %}
    {%- if attr is not empty %}
        {% for name, value in attr %}
            {{ ' ' ~ name }}={% if '"' in value %}'{{ value|raw }}'{% else %}"{{ value|raw }}"{% endif %}
        {% endfor %}
    {% endif %}
{% endmacro %}
```

Usage will be:

```twig
<div class="steps {{ _counter ? 'steps-counter' : '' }} {{ 'steps-'~_color }}" {{ attr_to_html(_attr) }}>
```

cavasinf:
Example with steps component.
Looks for ` # ---- HERE --- #}`

```twig
{% macro steps(steps, options) %}

    {% set _tooltip = options.tooltip ?? false %}
    {% set _color = options.color ?? 'primary' %}
    {% set _counter = options.counter ?? false %}
    {% set _attr = options.attr ?? {} %}

    <div 
           class="steps {{ _counter ? 'steps-counter' : '' }} {{ 'steps-'~_color }}" 
           {{ _self.attr_to_html(_attr) }}    {# ---- HERE --- #}
     > 

        {% set activeStepProcessed = false %}

        {% for step in steps %}
            {% set _type = (step.clickable ?? not activeStepProcessed) ? 'a' : 'span' %}
            {% set _title = step.title ?? '' %}
            {% set _description = step.description ?? _title %}
            {% set _active = step.active ?? false %}
            {% set _attr = step.attr ?? {} %}

            <{{ _type|raw }}
                class="step-item {{ _active ? 'active' : '' }}"
                href="{{ step.href ?? '#' }}"
                {%- if _tooltip -%} data-bs-toggle="tooltip" {%- endif -%}
                title="{{ _description }}"
[...]

PR #86, opened 2022-05-24
[Macros] Add `attr to html` twig macro for components

cavasinf:
## Description
Related to https://github.com/kevinpapst/TablerBundle/issues/85

## Types of changes

## Checklist

### DIFF
--- a/templates/includes/utils.html.twig
+++ b/templates/includes/utils.html.twig
@@ -3,7 +3,7 @@
 {% macro attr_to_html(attr) %}
     {%- if attr is not empty %}
         {% for name, value in attr %}
-            {{ name }}={{ value|e('html_attr') }}
+            {{ name }}="{{ value|e('html_attr') }}"
         {% endfor %}
     {% endif %}
 {% endmacro %}
--- a/templates/includes/utils.html.twig
+++ b/templates/includes/utils.html.twig
@@ -3,7 +3,7 @@
 {% macro attr_to_html(attr) %}
     {%- if attr is not empty %}
         {% for name, value in attr %}
-            {{ name|e('html_attr') }}={{ value|e('html_attr') }}
+            {{ name }}={{ value|e('html_attr') }}
         {% endfor %}
     {% endif %}
 {% endmacro %}
--- a/templates/includes/utils.html.twig
+++ b/templates/includes/utils.html.twig
@@ -1,0 +1,10 @@
+{# Twig used as macros holder only #}
+
+{% macro attr_to_html(attr) %}
+    {%- if attr is not empty %}
+        {% for name, value in attr %}
+            {{ name|e('html_attr') }}={{ value|e('html_attr') }}
+        {% endfor %}
+    {% endif %}
+{% endmacro %}
+
--- a/templates/components/button.html.twig
+++ b/templates/components/button.html.twig
@@ -1,4 +1,6 @@
 {% macro button(icon, values, type) %}
+    {% import '@Tabler/includes/utils.html.twig' as macros %}
+
     {%- apply spaceless -%}
         {% if type is same as (false) %}
             {# when using in dropdown actions for example, which might not have a "btn" class #}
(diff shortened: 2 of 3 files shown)


templates/includes/utils.html.twig
```diff
@@ -0,0 +1,10 @@
+{# Twig used as macros holder only #}
+
+{% macro attr_to_html(attr) %}
+    {%- if attr is not empty %}
+        {% for name, value in attr %}
+            {{ name|e('html_attr') }}={{ value|e('html_attr') }}
```
kevinpapst, line 6: are you missing the " here?

kevinpapst, line 6:
```suggestion
            {{ name|e('html_attr') }}="{{ value|e('html_attr') }}"
```

kevinpapst: Then we cannot use it on the attribute name.

But I am pretty sure that the escape filter does not add quotes, but only "fixes wrong character".

cavasinf:
```suggestion
            {{ name }}={{ value|e('html_attr') }}
```

cavasinf: resolved above

cavasinf:
Here is what does `e('html_attr')` twig filter

https://github.com/twigphp/Twig/blob/a1f8d3f91a0d28c258e81607e6711387a46f1705/src/Extension/EscaperExtension.php#L321

templates/includes/utils.html.twig
kevinpapst: I was 100% certain before that the escaper does not add quotes. 
And now that you linked to the code, I am even more certain :)

    {{ dump('ffooo') }}
    {{ dump('ffo"oo'|e('html_attr')) }}

Bildschirmfoto 2022-05-24 um 15 56 15

Did you check the generated HTML with the developer console, which shows the "tidy html" ?

cavasinf: Yes, 
I've already said that in https://github.com/kevinpapst/TablerBundle/issues/85#issuecomment-1135840318

Possible cases:
```html
<div data-good="good boy"></div>
```

```html
<div data-bad-question-mark="I'm not sure what I am doing.."></div>
```

@kevinpapst What did you expect to happen when using RAW HTML like that ?

cavasinf: Hey @kevinpapst, back from holidays!

Here is what you want: 

image


BUT this dump STRINGS not HTML.
I've done the same things for HTML render: 
image

Here's what is looks like in HTML (two last lines):
image

ATE, the render is the same 🤔

kevinpapst: Back from holidays and directly into such a discussion, I hope it doesn't lower your relaxation level 😁 

It is the same in your browser console, because every browser "fixes" invalid HTML as good as it can. And both the console and the element inspector will NOT show the original HTML which was sent to the browser, but the interpreted/cleaned version of it. 

The first version is not outputting the "" around the attributes value as we can see in the first screenshot. 

According to the RFC both versions are valid: 
https://www.w3.org/TR/2018/SPSD-html5-20180327/syntax.html#attributes-0

> There are four slightly different syntaxes that may be used for attributes in HTML: empty, unquoted, single-quoted and double-quoted. All four syntaxes may be used in the HTML syntax, depending on what is needed for each specific attribute. However, in the XHTML syntax, attribute values must always be quoted using either single or double quotes.

But I would still prefer if we render `Double-quoted attribute value syntax`, which is less error-prone than the `Unquoted attribute value syntax`.

templates/includes/utils.html.twig
```diff
@@ -0,0 +1,10 @@
+{# Twig used as macros holder only #}
+
+{% macro attr_to_html(attr) %}
+    {%- if attr is not empty %}
+        {% for name, value in attr %}
+            {{ name }}={{ value|e('html_attr') }}
```
cavasinf, line 6:
```suggestion
            {{ name }}="{{ value|e('html_attr') }}"
```

templates/includes/utils.html.twig
kevinpapst: Interesting are only:
```
    <div data-good={{ 'good boy' }}></div>
    <div data-good={{ 'good boy'|e('html_attr') }}></div>
```
The first one is wrong, becuase it renders as
    <div data-good=good boy></div> and the browser thinks `data-good=good` is a `Unquoted attribute value syntax` and `boy` is a `Empty attribute syntax`.

Now the same works when you attach the `|e('html_attr')` filter, because the space is escaped and therefor not interpreted as spacer for the next attribute.

Yes, in theory it should not matter if we add " or not, because the filter makes sure that there is no un-escaped space. But in practice it is better to add quotes, to be consistent everywhere.

Do NOT trust your browser console if you want to see the real TWIG output. I had some troublesome debugging sessions because of that misleading behavior! You donÄt see the problem, unless you dump the real HTML before sending it to the browser.

kevinpapst (approved):
Thanks 👍

Issue #89, opened 2022-05-31
Alert macro and `alert-important` class


cavasinf:
Why the class `alert-important` is automated when `title` or `description` are empty?

https://github.com/kevinpapst/TablerBundle/blob/0adfe1668516918f8fddb05f848bb28707e4a99e/templates/components/alert.html.twig#L3

It blocks me to add title + description AND full alert background color.
I don't think it is a good idea.

We can let this automation but add a new param `important` and change de background condition like
`{% set background = title is empty or description is empty or important ? true : false %}`

kevinpapst:
Your suggestion looks good to me 👍 

Probably clarify with `or important is same as (true)` otherwise it might fail if you pass `false`

Question: there are already too many params, shall we add an options array? We could do something like:
```twig
{% macro alert(type, description, title, icon, dismissible) %}
  {% if type is iterable %}
    {% set options = type %}
    {% set type = options.type %}
    {% set important = options.important %}
    ...
```
Then it would be possible to pass all values as one param array.

Or we build some helper macros that allow to call it with less params:

```twig
{% macro alert_important(description, title, icon) %}
  {{ self.alert('danger', description, title, icon, false, true) %}
```

cavasinf:
Due to:
- `alert-title` when title is set.
- `text-muted` when description is set.

There is no work around to manually add that class.
```twig
{{ alert('danger  alert-important', 'Vous allez clôturer le désordre : trigger ?' ,'Cloture du désordre', 'fas fa-exclamation-triangle') }}
```
Current:
![image](https://user-images.githubusercontent.com/25293190/171131218-eb6fc110-db91-46ba-b80b-cd30cb5c65d3.png)

Should be:
![image](https://user-images.githubusercontent.com/25293190/171131339-41e2682c-a0ea-4611-87fa-1f7af9225b57.png)

cavasinf:
> Question: there are already too many params, shall we add an options array?

Yeah, I've though the same thing..
Do you want to handle this macro params to not have BC break ?

As other components, they have only one `options` param.
And this allow us to add/modify code or params without breaking other dev macro usage.
But convert all current params to one param will BC break..

kevinpapst:
Yes, even though still in pre-release state, I would avoid BC breaks. 

Can you start with that? Untested and (be careful) typed here in Github 😁 

```twig
{% macro alert(type, description, title, icon, dismissible) %}
  {% if type is iterable %}
    {% set options = type %}
[...]

PR #90, opened 2022-05-31
Refactoring `alert()` macro

cavasinf:
## Description
Related to #89 

#### Usage

```twig
{% from '@Tabler/components/alert.html.twig' import alert %}

<div class="row">
    <div class="col-6">
        {{ alert('danger', 'Description' ,'Title', 'fas fa-exclamation-triangle', true) }}
        {{ alert('danger', 'Description' ,'Title', 'fas fa-exclamation-triangle') }}
        {{ alert('danger', 'Description' ,'Title', 'fas fa-exclamation-triangle') }} {# Error here #}
        {{ alert('warning', 'Description' ,'Title') }}
        {{ alert('warning', null ,'Title') }}
        {{ alert('warning', null ,'Title', 'fas fa-exclamation-triangle') }}
        {{ alert('success', 'Description') }}
        {{ alert('success', 'Description', null, 'fas fa-exclamation-triangle') }}
        {{ alert('primary') }}
    </div>

    <div class="col-6">
        {{ alert({title : 'Title V2', description: 'Description V2', icon : 'warning', dismissible : true}) }}
        {{ alert({title : 'Title V2', description: 'Description V2', icon : 'warning'}) }}
        {{ alert({title : 'Title V2', description: 'Description V2', icon : 'warning', important: true}) }} {# Looks good now #}
        {{ alert({title : 'Title V2', description: 'Description V2', type : 'warning'}) }}
        {{ alert({title : 'Title V2', type : 'warning'}) }}
        {{ alert({title : 'Title V2', icon : 'warning', type : 'warning'}) }}
        {{ alert({description : 'Description V2', type : 'success'}) }}
        {{ alert({description : 'Description V2', type : 'success', icon : 'warning'}) }}
        {{ alert({type : 'primary'}) }}
    </div>
</div>
```

#### Preview

image

## Types of changes

## Checklist

### DIFF
--- a/templates/components/alert.html.twig
+++ b/templates/components/alert.html.twig
@@ -37,10 +37,10 @@
             {% if _title is not empty or _description is not empty %}
                 <div>
                     {% if _title is not empty %}
-                        <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _title }}</h3>
+                        <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _title|raw }}</h3>
                     {% endif %}
                     {% if _description is not empty %}
-                        <p class="m-0">{{ _description }}</p>
+                        <p class="m-0">{{ _description|raw }}</p>
                     {% endif %}
                 </div>
             {% endif %}
--- a/templates/components/alert.html.twig
+++ b/templates/components/alert.html.twig
@@ -9,8 +9,7 @@
         {% set _icon = options.icon ?? null %}
         {% set _dismissible = (options.dismissible ?? false) is same as (true) %}
         {% set _important = (options.important ?? false) is same as (true) %}
-
-        {% set _background = _title is empty or _description is empty or _important %}
+        {% set _raw = (options.raw ?? false) is same as (true) %}
     {% else %}
         {% deprecated "Passing multiple parameters to 'alert()' Tabler macro is deprecated, use options object syntax." %}
 
@@ -19,9 +18,11 @@
         {% set _description = description is defined ? description|trans : null %}
         {% set _icon = icon ?? null %}
         {% set _dismissible = (dismissible ?? false) is same as (true) %}
-
-        {% set _background = _title is empty or _description is empty %}
+        {% set _important = false %}
+        {% set _raw = false %}
     {% endif %}
+
+    {% set _background = _title is empty or _description is empty or _important %}
 
     <div
             class="alert alert-{{ _type }} {% if _dismissible %}alert-dismissible{% endif %} {% if _background %}alert-important{% endif %}"
@@ -37,10 +38,10 @@
             {% if _title is not empty or _description is not empty %}
                 <div>
                     {% if _title is not empty %}
-                        <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _title|raw }}</h3>
+                        <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _raw ? _title|raw : _title }}</h3>
                     {% endif %}
                     {% if _description is not empty %}
-                        <p class="m-0">{{ _description|raw }}</p>
+                        <p class="m-0">{{ _raw ? _description|raw : _description }}</p>
                     {% endif %}
                 </div>
             {% endif %}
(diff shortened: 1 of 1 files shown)


templates/components/alert.html.twig
```diff
@@ -1,27 +1,51 @@
...
+    {% if type is iterable %}
+        {% set options = type %}
+
+        {% set _type = options.type ?? 'danger' %}
+        {% set _title = options.title ?? null %}
+        {% set _description = options.description ?? null %}
+        {% set _icon = options.icon ?? null %}
+        {% set _dismissible = (options.dismissible ?? false) is same as (true) %}
+        {% set _important = (options.important ?? false) is same as (true) %}
+
```
kevinpapst, line 12: {% set _raw = (options.raw ?? false) is same as (true) %}

templates/components/alert.html.twig
```diff
@@ -1,27 +1,51 @@
...
+                <div class="alert-icon">
+                    {{ tabler_icon(_icon, true) }}
+                </div>
+            {% endif %}
+
+            {% if _title is not empty or _description is not empty %}
                 <div>
-                    <i class="{% if background %}icon{% else %}alert-icon{% endif%} {{ icon|tabler_icon(false, icon) }}"></i>
+                    {% if _title is not empty %}
+                        <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _title|raw }}</h3>
```
kevinpapst, line 40: {% if _raw %}{{ _title|raw }}{% else %}{{ _title }}{% endif %}

kevinpapst (changes requested):
Can you make the raw filter optional by a setting like in my example.
This filter can easily lead to CSRF issues if you pass user generated values to it. 
Better activate it per case, instead of "blindly" using unescaped data.

cavasinf:
Everything's OK for me 👌

kevinpapst:
I will add your examples to the demo application. Thanks you @cavasinf 👍

cavasinf:
Yeah I haven't seen a doc for that kind of feature.
We can maybe add the doc of that component is this repo like others 🤔

kevinpapst:
This is what I see after the merge:
Bildschirmfoto 2022-06-01 um 09 38 27
Can you confirm that is still looks as in your screenshot?

kevinpapst:
> Yeah I haven't seen a doc for that kind of feature.

Because you still don't use the demo app, do you 😁

kevinpapst:
Okay, there is one BC break which is not matching the Tabler docs: https://preview.tabler.io/docs/alerts.html

Before:
Bildschirmfoto 2022-06-01 um 09 45 33
 
After:
Bildschirmfoto 2022-06-01 um 09 51 29

Could you have a look while I work on the docs?

cavasinf:
@kevinpapst Looks like it's coming from the `icon` class again.

Tabler user SVG, we use FontAwesome.
Can you try with `false` param at `tabler_icon` twig function.
https://github.com/kevinpapst/TablerBundle/blob/da494e86cab2d04f5ea500ab04850cb918d8a675/templates/components/alert.html.twig#L34

I remember it was "odd" without this class, but if it's okay for you, we can switch that 👍

cavasinf:
I'll check for title color + description => new PR ?

kevinpapst:
I don't mind the icon size/positioning, bigger looks better and I didn't even recognize it before you mentioned it ;-)

I was only talking about the colors, which worked before. New PR sounds good :)

kevinpapst:
Going through my deprecations log, I saw that we missed one: 
https://github.com/kevinpapst/TablerBundle/blob/main/templates/components/callout.html.twig#L3

`callout` is calling `alert` with single params 🙄

cavasinf:
What do you want to do ? 
Add the deprecated message and change multi params into object one param

kevinpapst:
Because important is the 6 param and it got lost somewhere in the refactoring process.
So have a consistent API we should change the signature as well:
```
{% from '@theme/components/alert.html.twig' import alert %}
{% if type is iterable %}
    {{ alert(type) }}
{% else %}
    {% deprecated "Passing multiple parameters to 'callout()' Tabler macro is deprecated, use options object syntax." %}
    {{ alert({type: type|default('danger'), description: description, title: title|trans, icon: icon, dismissible: false, important: true}) }}
{% else %}
```

What do you think ?

cavasinf:
Missing the trans of `description` (like OLD alert macro) but LGTM 👍

cavasinf:
- Missing the trans of `description` (like OLD alert macro)
- `icon` is required in callout macro ?
- `description` is required ?
- Maybe add `{% set options = type %}` inside if condition (if devs looks at code).

PR #91, opened 2022-06-03
BugFix Alert text style without `alert-important`

cavasinf:
## Description
Fix of https://github.com/kevinpapst/TablerBundle/pull/90

image

## Types of changes

## Checklist

### DIFF
--- a/templates/components/alert.html.twig
+++ b/templates/components/alert.html.twig
@@ -38,10 +38,18 @@
             {% if _title is not empty or _description is not empty %}
                 <div>
                     {% if _title is not empty %}
-                        <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _raw ? _title|raw : _title }}</h3>
+                        {% if _background %}
+                            <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _raw ? _title|raw : _title }}</h3>
+                        {% else %}
+                            <h4 class="alert-title">{{ _raw ? _title|raw : _title }}</h4>
+                        {% endif %}
                     {% endif %}
                     {% if _description is not empty %}
-                        <p class="m-0">{{ _raw ? _description|raw : _description }}</p>
+                        {% if _background %}
+                            <p class="m-0">{{ _raw ? _description|raw : _description }}</p>
+                        {% else %}
+                            <div class="text-muted">{{ _raw ? _description|raw : _description }}</div>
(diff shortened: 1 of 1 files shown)


templates/components/alert.html.twig
```diff
@@ -38,10 +38,18 @@
             {% if _title is not empty or _description is not empty %}
                 <div>
                     {% if _title is not empty %}
-                        <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _raw ? _title|raw : _title }}</h3>
+                        {% if _background %}
+                            <h3 class="{{ _description is not empty ? 'mb-1' : 'mb-0' }}">{{ _raw ? _title|raw : _title }}</h3>
```
kevinpapst, line 42: why is it getting bigger with background?

templates/components/alert.html.twig
```diff
@@ -38,10 +38,18 @@
...
+                        {% else %}
+                            <h4 class="alert-title">{{ _raw ? _title|raw : _title }}</h4>
+                        {% endif %}
                     {% endif %}
                     {% if _description is not empty %}
-                        <p class="m-0">{{ _raw ? _description|raw : _description }}</p>
+                        {% if _background %}
+                            <p class="m-0">{{ _raw ? _description|raw : _description }}</p>
```
kevinpapst, line 49: I think it is better to keep the same elements `h4` and `div` and only remove the classes if background is active

templates/components/alert.html.twig
cavasinf: Because Tabler doesn't handle title/description content as the sames way if `alert-important` or not.

#### normal : h4 + div
image

#### `alert-important` : d-flex + divs
image

templates/components/alert.html.twig
cavasinf: Linked to above reply

templates/components/alert.html.twig
kevinpapst: Hm, annoying that we cannot re-use the same structure. 
Then there are 8 flows:
- important, title, title + description, description
- not important, title, title + description, description
and each has a different structure?

cavasinf: As soon Tabler uses icon or other (in column) with alerts, they add `d-flex` container.
image

That why I've choose to "getting bigger with background" instant of a new whole block.

kevinpapst (approved):
LGTM , thank you 👍

PR #93, opened 2022-06-09
Fix callout embed

kevinpapst:
## Description

Needed changes after https://github.com/kevinpapst/TablerBundle/pull/90

## Types of changes

## Checklist

### DIFF
--- a/templates/components/callout.html.twig
+++ b/templates/components/callout.html.twig
@@ -1,4 +1,17 @@
 {% macro callout(type, description, title, icon) %}
-    {% import '@Tabler/components/alert.html.twig' as macro %}
-    {{ macro.alert(type|default('danger'), description, title, icon, false, true) }}
+    {% from '@Tabler/components/alert.html.twig' import alert %}
+    {% if type is iterable %}
+        {% set type = {dismissible: false, important: true}|merge(type) %}
+        {{ alert(type) }}
+    {% else %}
+        {% deprecated "Passing multiple parameters to 'callout()' Tabler macro is deprecated, use options object syntax." %}
+        {{ alert({
+            type: type,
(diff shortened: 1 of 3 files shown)


templates/components/callout.html.twig
```diff
@@ -1,4 +1,17 @@
...
+    {% from '@Tabler/components/alert.html.twig' import alert %}
+    {% if type is iterable %}
+        {% set type = {dismissible: false, important: true}|merge(type) %}
+        {{ alert(type) }}
+    {% else %}
+        {% deprecated "Passing multiple parameters to 'callout()' Tabler macro is deprecated, use options object syntax." %}
+        {{ alert({
+            type: type,
+            description: (description is null ? null : description|trans),
+            title: (title is null ? null : title|trans),
```
cavasinf, line 11: `is null` and `is defined` do the same job ?

What if:
- `callout('danger')`
- `callout('danger', null, null, 'warning')`
- `callout('danger', undefined, undefined, 'warning')`

kevinpapst: These variables are always defined, because they are the macro function params.
But just to be sure I dumped them into the demo app and here is the result of these strange use-cases:
Bildschirmfoto 2022-06-09 um 14 54 01

Are you writing a lot of javascript lately? There is no `undefined` in twig 😁

cavasinf: Yeah I'm working on React Native lately 😭

How do you want to handle not defined/null params ?
1. xxxx is defined ? true : false
2. xxxx is null ? false : true
3. xxxx is not null ? true : false

cavasinf: OK 👍

cavasinf (commented):
LGTM 👍

PR #97, opened 2022-06-27
removed unused and some very opinionated code from the forms theme

kevinpapst:
## Description

This is a BC break if you used the form these provided by this bundle.
Otherwise it is just a cleanup of code that does not belong here.

## Types of changes

## Checklist


kevinpapst:
The code which I removed should not be in here, it is the responsibility of the project to decide how your forms should look like. Floating labels work different from regular forms. Then also there were leftovers from certain datepicker-picker libraries in there, which should not interfere with your customizations.

@wucherpfennig @cavasinf can you have a quick look if your projects are negatively affected / if the docs are sufficient to clarify the process.

wucherpfennig:
Sorry for not providing feedback. I did check and it seams reasonable. I will find out as soon as there is an obvious error aka "scream testing" 😉

PR #98, opened 2022-07-01
New status() and status_dot() macros

wucherpfennig:
## Description
Added `status()` and `status_dot` macros

## Types of changes

## Checklist

### DIFF
--- a/templates/components/status.html.twig
+++ b/templates/components/status.html.twig
@@ -1,0 +1,16 @@
+{% macro status(text, options) %}
+    {% set _color = options.color ?? 'green' %}
+    {% set _animated = options.animated ?? true %}
+    {% set _lite = options.lite is defined ? options.lite  : false %}
+    {% set _with_dot = options.with_dot is defined ? options.with_dot  : false %}
+    {% set _extraClass = options.extraClass ?? '' %}
+
+    {%- apply spaceless -%}
+        <span class="status status-{{ _color }} {{ _lite ? 'status-lite' : '' }} {{ _extraClass }}">
+        {% if _with_dot %}
+            <span class="status-dot {{ _animated ? 'status-dot-animated' : '' }}"></span>
+        {% endif %}
+            {{ text }}
+        </span>
+    {%- endapply -%}
+{% endmacro %}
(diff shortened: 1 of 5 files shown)


templates/components/status.html.twig
```diff
@@ -0,0 +1,16 @@
+{% macro status(text, options) %}
+    {% set _color = options.color ?? 'green' %}
+    {% set _animated = options.animated ?? true %}
```
kevinpapst, line 3: does that really work? I always had in mind that this will fail if you pass `false` - as in the next two lines. Or does it only fail if you set `true` as default param and want to pass `false`?

wucherpfennig: Yes it does ;-) at least this is how I am using it. I have attached as screenshot with some "tests".

`??` does check for `isset() && !is_null()`  hence this "shorthand" notation. Therefore you can pass `FALSE` without any problems. [(Stackoverflow)](https://stackoverflow.com/questions/53610622/what-does-double-question-mark-operator-mean-in-php)

```twig
    {% set options = {'color': 'success', 'with_dot': true,'animated': true, 'lite': false, 'extraClass': 'me-3'} %}
    {{ status('Demo 1', options) }}

    {% set options = {'color': 'info', 'with_dot': false,'animated': true, 'lite': true, 'extraClass': 'me-3'} %}
    {{ status('Demo 2', options) }}

    {% set options = {'color': 'orange', 'with_dot': true,'animated': true, 'lite': false, 'extraClass': 'me-3'} %}
    {{ status('Demo 3', options) }}

    {% set options = {'color': 'pink', 'extraClass': 'me-3'} %}
    {{ status('Demo 4', options) }}

    {{ status('Demo 5') }}

    {{ status('Demo 6', false) }}

    {{ status('Demo 7', null) }}
```

grafik

kevinpapst: Thanks, and I I will use those for the demo app 👍

kevinpapst (approved):

kevinpapst:
Thanks you @wucherpfennig for your continuous contributions 👍 

I updated the demo app and added examples for all possible status/status_dot/status_indicator combinations

PR #107, opened 2022-09-08
Update components-accordion.md

cavasinf:
Just saw that camelCase error 😭


kevinpapst (approved):
good spotting, missed it

PR #113, opened 2022-09-16
BugFix themes link

cavasinf:
## Description
Fix themes link in doc

## Types of changes

## Checklist

### DIFF
--- a/docs/form_theme.md
+++ b/docs/form_theme.md
@@ -1,9 +1,9 @@
 # Form theme
 
 This bundle does ship three form-themes:
-- [templates/layout/form-theme.html.twig](templates/layout/form-theme.html.twig)
-- [templates/layout/form-theme-horizontal.html.twig](templates/layout/form-theme-horizontal.html.twig)
-- [templates/layout/form-theme-security.html.twig](templates/layout/form-theme-security.html.twig)
+- [templates/layout/form-theme.html.twig](/templates/layout/form-theme.html.twig)
+- [templates/layout/form-theme-horizontal.html.twig](/templates/layout/form-theme-horizontal.html.twig)
+- [templates/layout/form-theme-security.html.twig](/templates/layout/form-theme-security.html.twig)
 
 None of these themes is automatically loaded, you have to use it manually.
 


docs/form_theme.md
```diff
@@ -1,9 +1,9 @@
 # Form theme
 
 This bundle does ship three form-themes:
-- [templates/layout/form-theme.html.twig](templates/layout/form-theme.html.twig)
-- [templates/layout/form-theme-horizontal.html.twig](templates/layout/form-theme-horizontal.html.twig)
-- [templates/layout/form-theme-security.html.twig](templates/layout/form-theme-security.html.twig)
+- [templates/layout/form-theme.html.twig](/templates/layout/form-theme.html.twig)
```
kevinpapst, line 4:
```suggestion
- [templates/layout/form-theme.html.twig](https://github.com/kevinpapst/TablerBundle/blob/main/templates/layout/form-theme.html.twig)
```

kevinpapst (changes requested):
I'd say the only way to make it universally working is to use absolut links?!

docs/form_theme.md
cavasinf: I don't think so.

If in `main` branch:
- `/xxxx` means `https://github.com/kevinpapst/TablerBundle/blob/main/xxxx`
- `xxxx` means `https://github.com/kevinpapst/TablerBundle/blob/main/{{pathToCurrentFile}}/xxxx`

If we especially prefix links with `https://github.com/kevinpapst/TablerBundle/blob/main`,
that means every forks of the bundle with any branches will redirect to `main` from here. 
It does not matter which depot or branch you are in.

kevinpapst: I do not care about forks... and who is reading branch docs? 

Anyway:

> * `/xxxx` means `https://github.com/kevinpapst/TablerBundle/blob/main/xxxx`

Is that true? Is Github dynamically replacing these links?
I learned something new today and your solution looks good then!

kevinpapst (approved):

docs/form_theme.md
cavasinf: > I do not care about forks... 

I'm in that case for FOSUserBundle, forked it to not have swift mailer dependency but still have doc updated.

> and who is reading branch docs?

Some PR will update docs, new component/macro/..
In PR user side, his links should point to `main` branch ?
Kinda weird to me.

> Is that true? Is Github dynamically replacing these links?

Yeah, you can check on main branch now it's merged 😉 
https://github.com/kevinpapst/TablerBundle/blob/main/docs/form_theme.md

kevinpapst: That was more of a rhetorical question, I tried your changed file directly in the PR via "View file" and it worked 😁 so I knew you were right. I just didn't know that or just forgot about that feature... but it makes writing docs much easier, which is really good to know!

PR #115, opened 2022-09-22
Update theme override doc

cavasinf:
## Description
Add more detailed DOC about theme override

## Types of changes

## Checklist

### DIFF
--- a/docs/theme.scss
+++ b/docs/theme.scss
@@ -111,10 +111,29 @@ h1, h2, h3, h4, h5, h6 {
   font-family: Montserrat, var(--tblr-font-sans-serif); // font only for titles
 }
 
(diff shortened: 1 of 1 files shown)


kevinpapst:
@cavasinf you write "add more detailed doc" but the PR includes only the adapted scss file.
Can you share more details? Because I do not get it. Is this your set of fixes for Tabler? What does it do?

cavasinf:
I've just added more scss color override like `light backgrounds, `border colors` and `text colors`.
They were not overridden by variables colors of Tabler's default css.

I mean `doc` because it is not used by the Bundle. It's just an example if devs want to change default Tabler's colors.

kevinpapst (approved):

kevinpapst:
Okay then 😄

PR #118, opened 2022-09-30
Add wizard "template"

cavasinf:
## Description
Add wizard telplate: https://preview.tabler.io/wizard.html

⚠️ Extends from `fullpage` template

## Types of changes

## Checklist


kevinpapst (commented):
Any reason why all blocks are named `wizzard` and not `wizard`?

cavasinf:
Because of "blizzard" ? 🤔

No I've just messed up this part..

templates/wizard.html.twig
```diff
@@ -0,0 +1,47 @@
...
+            {% block container_attr %}{% endblock %}
+    >
+        {% block wizzard_before %}{% endblock %}
+
+        {% block wizzard_logo %}
+            <div class="text-center mb-4">
+                <h1 class="navbar-brand navbar-brand-autodark mb-4">
+                    <a href="{{ path('tabler_welcome'|tabler_route) }}">
+                        {% if tabler_bundle.getLogoUrl() is not empty %}
+                            <img src="{% if '://' in tabler_bundle.getLogoUrl() %}{{ tabler_bundle.getLogoUrl() }}{% else %}{{ asset(tabler_bundle.getLogoUrl()) }}{% endif %}"
```
kevinpapst, line 16: Having an empty `alt=""` might be better than a default string

kevinpapst: Yes, good idea!

cavasinf: Logo code is now under `embed` => [/templates/embeds/logo.html.twig]()

templates/layout-horizontal.html.twig
```diff
@@ -35,15 +35,9 @@ Enjoy your theme!
...
-                    <h1 class="navbar-brand d-none-navbar-horizontal pe-0 pe-md-3{% block navbar_brand_class %} navbar-brand-autodark{% endblock %}">
-                        <a href="{{ path('tabler_welcome'|tabler_route) }}">
-                            {% if tabler_bundle.getLogoUrl() is not empty %}
-                                <img src="{% if '://' in tabler_bundle.getLogoUrl() %}{{ tabler_bundle.getLogoUrl() }}{% else %}{{ asset(tabler_bundle.getLogoUrl()) }}{% endif %}" width="110" height="32" alt="Logo" class="navbar-brand-image">
-                            {% else %}
-                                <i class="far fa-smile" style="font-size: 32px"></i>
-                            {% endif %}
-                        </a>
-                    </h1>
+                    {% embed '@Tabler/embeds/logo.html.twig' %}
```
kevinpapst, line 38: Have you tried if it is possible to override `navbar_brand_classes` in your project?  I assume not and this would be a BC break. 

Shall we probably only put the inner `{% if tabler_bundle.getLogoUrl() is not empty %}` into the embed and leave the `h1` in here?

cavasinf: Ahhhh... Here we go again.. I forget that but we cannot override block of an embed..
Should I revert my last commit ?

kevinpapst: Either that or we move only this part to the embed, leaving the h1 in the original layouts
```
{% if tabler_bundle.getLogoUrl() is not empty %}
                                <img src="{% if '://' in tabler_bundle.getLogoUrl() %}{{ tabler_bundle.getLogoUrl() }}{% else %}{{ asset(tabler_bundle.getLogoUrl()) }}{% endif %}" width="110" height="32" alt="Logo" class="navbar-brand-image">
                            {% else %}
                                <i class="far fa-smile" style="font-size: 32px"></i>
                            {% endif %}
```

cavasinf: 👍

kevinpapst (commented):
Two tiny things I'd like to mention: 

1. I tried to setup a demo of the wizard and would say that this template does not help that much. Shall we probably add some macros to render e.g. progress and button?
2. The filename: shall we call it `layout-wizard` ?

cavasinf:
1. Good idea, to replicate the Tabler demo page structure https://preview.tabler.io/wizard.html with two buttons bottom right + progress bar. TBH this is what I've done to my wizard template on my app 😉 
2. IDK.. In that case, what do we do with `fullpage.html.twig` ?

kevinpapst:
1. See, same in the demo. Do you want to do it before merging?
2. Hm, missed that. Next bundle I start, I will think about naming before^^ then leave as is. The ones with `layout-` are the real layouts and the other ones and more like helper for special occasions.

templates/wizard.html.twig
```diff
@@ -0,0 +1,63 @@
...
+                            {{ button(false, { title : 'Close'|trans({}, 'TablerBundle'), combined: true, url : '#' }, 'link') }}
+                        {% endblock %}
+
+                        {% block next_button %}
+                            {{ submit_button(false, { title : 'Show homepage'|trans({}, 'TablerBundle'), combined: true }, 'primary') }}
+                        {% endblock %}
+                    </div>
+                </div>
+            </div>
+        {% endblock %}
```
cavasinf, line 56: 1.Do we care about default block name ? Prefix with `wizard` ?

kevinpapst: Ah, after reading it 5 times, I get it... trans != trad
Yeah, why not adding new translations. 
I will add Weblate support, so the community can translate them.

First I was thinking if we should add some variables, eg. `progress_current` or `back_url` but I guess we have to override the blocks in the project anyway, so that would not cover much more use-cases ... but I have no immediate use-case and find it hard to imagine. 
But yes, my gut says add `wizard_` for these blocks.

cavasinf: OK, I'll rename `back` to `previous` and add trans + prefix block names.
I can only do that next Monday though.

cavasinf: Good to go 👍

kevinpapst (approved):
Thank you @cavasinf - if you feel like documentation 😁 please raise another PR.

PR #121, opened 2022-10-04
Move Twig block "body_class" into the class attribute of BODY

Etrimi:
## Description
If we add class in the Twig block "body_class", the text is put AFTER the value of the class attribute.
Ex: 
`{% block body_class %}layout-boxed{% endblock %}`

Result:
`<body class="antialiased" layout-boxed>`

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [X] I agree that this code will be published under the [MIT license](https://github.com/kevinpapst/TablerBundle/blob/master/LICENSE)

### DIFF
--- a/templates/layout-horizontal.html.twig
+++ b/templates/layout-horizontal.html.twig
@@ -23,7 +23,7 @@ Enjoy your theme!
         {% endif %}
     {% endblock %}
 </head>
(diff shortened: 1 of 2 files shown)


kevinpapst (approved):
Looks like a copy & paste mistake fron the past. Thanks @Etrimi 👍

PR #122, opened 2022-10-08
include for html_head and rel=nofollow

kevinpapst:
## Description
- extract html head default tags to include (allows to override via new file in app `templates/bundles/TablerBundle/`templates/includes/html_head.html.twig`)
- add `rel=noindex.nofollow` to prevnt public apps from showing up in search engines

## Types of changes

## Checklist


cavasinf:
are you sure about the default "robots" meta?

kevinpapst:
What do you mean by "sure"? 
Obviously, otherwise I would not have added it 😁 
I am sure that most people will use this bundle for internal apps and those should not appear in search engines.

Some companies like to link their internal tools from their homepage, so they can be found easier by employees.
And you can still override that tag if you want to. I just want the defaults to be strict.

cavasinf:
OK for me then 👍

PR #123, opened 2022-10-18
update to beta 13

kevinpapst:
## Description

Want to give it a try @cavasinf @wucherpfennig before merge?

## Types of changes

## Checklist


kevinpapst:
we have to update anyway at some point, so do it now 🚀

cavasinf:
Yeah until V1, I don't think we need to test every Tabler update.
Of course, there will be some errors, but it's part of using a beta Template.

kevinpapst:
Honestly, a part of me does not believe that there will be a stable v1 ever.
Feels more like feature creep in the middle of a beta cycle.
I do not see a stabilizing phase or even a feature freeze coming up. But I'd be glad to be proven wrong.

cavasinf:
FYI
> error @tabler/core@1.0.0-beta13: The engine "node" is incompatible with this module. Expected version ">=16". Got "14.19.1"

Had to downgrade to `beta12`

kevinpapst:
Edit package.json, replace beta13/14 with beta12, yarn install, yarn run build.

But why would you do that? There is no way backward 😁 

And the latest release is really the best one ever. I have literally no bug and no workaround SCSS locally

kevinpapst:
Oh, okay. My stack was more up-to-date I guess. Sounds like a big chunk of work.

I would always build locally and commit the new assets. 
Deploy only what was checked by some human.
Why would you let a CI process do that? Sounds risky...

Anyway, now you are on recent versions and running beta14?

cavasinf:
> Sounds like a big chunk of work.

True!

> I would always build locally and commit the new assets.

I build locally to dev AND from CI/CD to deploy

> Why would you let a CI process do that? Sounds risky...

I have multiple jobs done by my runner:
- php dependencies checks (without dev mode)
- lint
- yarn package deprecated/vulnerability checks
- yarn build
- Symfony security check
- Doctrine Migrations check

If one of those fail, deployment will not be proceed.

ATE, to deploy my project with the asset, my runner upload final builded files as artifacts to the server, it allows me to get rid of Node.js on our production servers: https://symfony.com/doc/current/frontend/encore/faq.html#how-do-i-deploy-my-encore-assets

PR #128, opened 2022-12-07
Added null check to timeline item parts

wucherpfennig:
## Description
it seems useful to me...

## Types of changes

## Checklist

### DIFF
--- a/templates/components/timeline.html.twig
+++ b/templates/components/timeline.html.twig
@@ -18,7 +18,7 @@
                             <p class="list-timeline-title">{{ item.title }}</p>
                         {% endif %}
                         {% if item.text is defined and item.text is not null %}
-                            <p class="text-muted">{{ item.text|raw }}</p>
+                            <p class="text-muted">{{ item.text }}</p>
                         {% elseif item.textHtml is defined and item.textHtml is not null %}
                             {{ item.textHtml|raw }}
                         {% endif %}
--- a/templates/components/timeline.html.twig
+++ b/templates/components/timeline.html.twig
@@ -11,15 +11,15 @@
                 {% endif %}
                 {% if true %}
                     <div class="list-timeline-content">
-                        {% if item.time is defined %}
+                        {% if item.time is defined and item.time is not null %}
                             <div class="list-timeline-time">{{ item.time }}</div>
                         {% endif %}
-                        {% if item.title is defined %}
+                        {% if item.title is defined and item.title is not null %}
                             <p class="list-timeline-title">{{ item.title }}</p>
                         {% endif %}
-                        {% if item.text is defined %}
+                        {% if item.text is defined and item.text is not null %}
                             <p class="text-muted">{{ item.text|raw }}</p>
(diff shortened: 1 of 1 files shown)


templates/components/timeline.html.twig
```diff
@@ -11,15 +11,15 @@
...
+                        {% if item.time is defined and item.time is not null %}
                             <div class="list-timeline-time">{{ item.time }}</div>
                         {% endif %}
-                        {% if item.title is defined %}
+                        {% if item.title is defined and item.title is not null %}
                             <p class="list-timeline-title">{{ item.title }}</p>
                         {% endif %}
-                        {% if item.text is defined %}
+                        {% if item.text is defined and item.text is not null %}
                             <p class="text-muted">{{ item.text|raw }}</p>
```
kevinpapst, line 21: I know you did not change it, but while looking at your changes I saw that line as well.
What do you think? In my opinion the |raw should be removed here. It is dangerous if used in a context where the user defined the text.

wucherpfennig: imho: `item.text` should be `plain text`. otherwise one could use `item.textHtml`

What I did not check is whether the above changes 'destroy' the template if you do not have eg. `item.time` set.
I will check on that later.

cavasinf: ➕ 1 to remove the `raw` with the simple `text` option.

kevinpapst (approved):

kevinpapst:
Thanks @wucherpfennig 👍 

I did a small adjustment as well and made the time a mandatory field to render an item

cavasinf:
ATE we do not allows item if there's no `time` ? Is that a good idea ?

I use that component as a vertical "configurator stepper" or "status indicator".
I have no time to put in there and I'm obliged to put an empty string to each item now ?

image

templates/components/timeline.html.twig
```diff
@@ -9,17 +9,15 @@
                         {{ tabler_icon(item.icon, false) }}
                     </div>
                 {% endif %}
-                {% if true %}
+                {% if if item.time is defined and item.time is not null %}
```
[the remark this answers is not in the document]
kevinpapst: Thanks for finding, fixed in master!

I thought a twig:lint is run here, but this is "only" a bundle.

kevinpapst:
A timeline should have a time, don't you think so? 

If you use it differently, which is legitimate, then we should add a wrapper function that adds empty time attributes. Or send a PR to revert that change and make it better 👍

kevinpapst:
There is a `_isSimple` mode, I added a check for that, see https://github.com/kevinpapst/TablerBundle/commit/c3dc610d3d36f6fd5d4c7b12c3f2f7b0a3d339a0

cavasinf:
As Tabler 16 implemented an improved stepper : https://preview.tabler.io/steps.html
There's no need to do that anymore. 

Need to implements `vertical` to the [steps component](https://github.com/kevinpapst/TablerBundle/blob/main/templates/components/steps.html.twig)

PR #130, opened 2022-12-23
[Card vertical] BugFix file name

cavasinf:
## Description
Oops 🤫 

## Types of changes

## Checklist


kevinpapst (approved):

cavasinf:
We do that with our `bundle` which is imported from a sub-module.
But I don't know how to do that with composer/vendor import 🤔

kevinpapst:
Now that we write ... maybe with a complete new workflow, which runs "composer create ...new-symfony-app..." and then "composer require git@github.com:kevinpapst/TablerBundle.git".

PR #143, opened 2023-02-09
[Login] Add Login-cover page

cavasinf:
## Description
Add login cover page from Tabler : https://preview.tabler.io/sign-in-cover.html

<del>

⚠️ Many parts of the code is copy-pasta from `security.html.twig`.
The only "new" code is before the login form AND after.

As a twig "way to work", devs cannot override blocks inside of an embed/include..
How do you want to handle that duplicated code @kevinpapst?

</del>

Extended from `/security.html.twig`

## Types of changes

## Checklist

### DIFF
--- a/docs/layout.md
+++ b/docs/layout.md
@@ -32,7 +32,7 @@ The `security.html.twig` file for login, register ... can be used with:
 
 See example at [https://preview.tabler.io/sign-in.html](https://preview.tabler.io/sign-in.html)
 
-**For your security screens (with huge cover image**
+**For your security screens (with huge cover image)**
 
 The `security-cover.html.twig` file for login, register ... can be used with:   
 ```
--- a/docs/layout.md
+++ b/docs/layout.md
@@ -34,7 +34,7 @@ See example at [https://preview.tabler.io/sign-in.html](https://preview.tabler.i
 
 **For your security screens (with huge cover image)**
 
-The `security-cover.html.twig` file for login, register ... can be used with:   
+The `security-cover.html.twig` file (for login, register, ...) can be used with:   
 ```
 {% extends '@Tabler/security-cover.html.twig' %}
 ```
--- a/docs/layout.md
+++ b/docs/layout.md
@@ -30,6 +30,19 @@ The `security.html.twig` file for login, register ... can be used with:
 {% extends '@Tabler/security.html.twig' %}
 ```
 
+See example at [https://preview.tabler.io/sign-in.html](https://preview.tabler.io/sign-in.html)
+
+**For your security screens (with huge cover image**
+
+The `security-cover.html.twig` file for login, register ... can be used with:   
+```
+{% extends '@Tabler/security-cover.html.twig' %}
+```
+
+The cover image is configurable at `tabler.options.security_cover_url` or via `ConextHelper::setSecurityCoverUrl()`.
+
+See example at [https://preview.tabler.io/sign-in-cover.html](https://preview.tabler.io/sign-in-cover.html)
+
 **For your error pages**
 
 The `error.html.twig` for all of your [error pages](error_pages.md):   
--- a/templates/security-cover.html.twig
+++ b/templates/security-cover.html.twig
@@ -100,7 +100,7 @@
 
     <div class="col-12 col-lg-6 col-xl-8 d-none d-lg-block">
         {% block before_cover %}{% endblock %}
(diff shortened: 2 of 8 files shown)


kevinpapst:
I copy & pasted the layout into my IDE to see the differences and at a first glance I would say we can handle that all in one layout with more blocks.

Then we can simply document how to achieve this new layout, which is currently missing the social (oauth) login section.

The only tricky part is the `div class="card"`, which is not needed in this new layout. Maybe we can leave that and simply remove the class?

cavasinf:
> I would say we can handle that all in one layout with more blocks

I'm scared of blocks inside includes/embeds..

So only a conditional variable from the extend should be enough?
We can go with the same concept for the vertical/horizontal menu condition with `layout_type`

https://github.com/kevinpapst/TablerBundle/blob/2d3213ffe4c1cb7b689797590c9582315c1fe0eb/templates/layout-vertical.html.twig#L49

kevinpapst:
> I'm scared of blocks inside includes/embeds..

I don't think (or at least I hope) we need that. I am more scared of code duplication.

Can't we just add some more blocks inside the existing layout file? E.g. for the body class.

cavasinf:
> Can't we just add some more blocks inside the existing layout file? E.g. for the body class.

Sure, I'll do it when I have the time to 😉

kevinpapst (approved):
I did some minor adjustments, please check/test and then lets merge.

Seems to be easier to have two layout files instead of complicated block setups.

kevinpapst (changes requested):
Oh, two minor things:
- Let's move it to `templates/security-cover.html.twig` and 
- add a sentence in the docs?

cavasinf:
> Let's move it to `templates/security-cover.html.twig`

OK but this means that developers can use this layer as a `forget password`? (for example)

> add a sentence in the docs?

Always 😉

kevinpapst:
> OK but this means that developers can use this layer as a forget password? (for example)

I hope so, even though I haven't tested it. 

We should make the image configurable through the context object.
Will push some updates later on as well.

kevinpapst:
Please review last commit: moved new layout file, added configuration for cover url, added docs

templates/security-cover.html.twig
```diff
@@ -1,112 +1,19 @@
...
-                    {% endif %}
-                {% endblock %}
-            {% endblock %}
-        </div>
-    </div>
+
+    {% block container_classes %}container container-tight my-5 px-lg-5{% endblock %}
+    {% block card_classes %}{% endblock %}
+
+    {% block container_after %}
```
kevinpapst, line 11: that doesn't compile

cavasinf:
Tested in Demo by only replacing the extend in `login.html.twig`, 
from `{% extends '@Tabler/security.html.twig' %}`
to `{% extends '@Tabler/security-cover.html.twig' %}`

Capture d’écran 2023-02-20 à 10 27 06

templates/security-cover.html.twig
cavasinf: WDYM?

kevinpapst: You open a block and close a div: that feels / looks wrong 😁

Maybe it is the indentation that is actually the problem. 
Is `container_after` closed in line 20?

cavasinf: Yes, `div` is opened with `block container_before` at L5
and closed at L12 with  `block container_after`.

The indents are like that to remind me that a DIV is open and not closed.
If the end of my file has one (or more) TABs, it means something is wrong.

docs/layout.md
```diff
@@ -30,6 +30,19 @@ The `security.html.twig` file for login, register ... can be used with:
 {% extends '@Tabler/security.html.twig' %}
 ```
 
+See example at [https://preview.tabler.io/sign-in.html](https://preview.tabler.io/sign-in.html)
+
+**For your security screens (with huge cover image**
```
cavasinf, line 35:
```suggestion
**For your security screens (with huge cover image)**
```

docs/layout.md
```diff
@@ -30,6 +30,19 @@ The `security.html.twig` file for login, register ... can be used with:
...
+
+The `security-cover.html.twig` file for login, register ... can be used with:
```
cavasinf, line 37:
```suggestion
The `security-cover.html.twig` file (for login, register, ...) can be used with:   
```

kevinpapst (approved):
Awesome ❤️ seems to be BC safe and works in the demo!.

Thank you @cavasinf 👍

PR #155, opened 2023-03-31
[Card] Add `.card-body` print condition

cavasinf:
## Description

A Card with a fullsized table from Tabler should apparently not be inside `card-body`.

See here for Tabler demo: https://preview.tabler.io/tables.html (see `invoices` table)
image

With empty body and no padding (`fullsize` card template option) + Table inside card-body
image

I've added condition to not always print this part.

## Types of changes

## Checklist

### DIFF
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -21,7 +21,7 @@
 {% set _hasTitle            = title is not empty %}
 {% set _hasTools            = tools is not empty %}
 {% set _hasHeader           = header is not empty %}
-{% set _hasBody             = body is not empty or _force_body %}
+{% set _hasBody             = _use_body or body is not empty %}
 {% set _hasFooter           = _use_footer or footer is not empty %}
 
 {% import '@Tabler/components/buttons.html.twig' as button %}
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -8,6 +8,7 @@
 {% set _collapsible         = collapsible ?? _collapsed %}
 {% set _footer_collapsible  = footer_collapsible ?? true %}
 {% set _use_footer          = use_footer ?? false %}
+{% set _force_body          = force_body ?? false %}
 
 {# Common blocks : Declared this way to use `is not empty` and prevent the `has already been rendered` form field twig error #}
 {% set header %}{% block box_header %}{% endblock %}{% endset %}
@@ -20,7 +21,7 @@
 {% set _hasTitle            = title is not empty %}
 {% set _hasTools            = tools is not empty %}
 {% set _hasHeader           = header is not empty %}
-{% set _hasBody             = body is not empty %}
+{% set _hasBody             = body is not empty or _force_body %}
 {% set _hasFooter           = _use_footer or footer is not empty %}
 
 {% import '@Tabler/components/buttons.html.twig' as button %}
@@ -59,9 +60,11 @@
 
     {# Body #}
     {% block box_body_before %}{% endblock %}
+    {% if _hasBody %}
     <div class="{% block box_type_class %}card-body {% endblock %}{% block box_body_class %}{% endblock %} {% if _fullsize %}p-0{% endif %} {% if _collapsible %} {{ _collapsible_class ~ (_collapsed ? ' collapse' : ' show') }} {% endif %}">
         {{ body|raw }}
     </div>
+    {% endif %}
     {% block box_body_after %}{% endblock %}
 
     {# Footer #}
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -8,7 +8,7 @@
 {% set _collapsible         = collapsible ?? _collapsed %}
 {% set _footer_collapsible  = footer_collapsible ?? true %}
 {% set _use_footer          = use_footer ?? false %}
-{% set _use_body          = use_body ?? true %}
+{% set _use_body            = use_body ?? true %}
 
 {# Common blocks : Declared this way to use `is not empty` and prevent the `has already been rendered` form field twig error #}
 {% set header %}{% block box_header %}{% endblock %}{% endset %}
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -8,7 +8,7 @@
 {% set _collapsible         = collapsible ?? _collapsed %}
 {% set _footer_collapsible  = footer_collapsible ?? true %}
 {% set _use_footer          = use_footer ?? false %}
-{% set _force_body          = force_body ?? true %}
+{% set _use_body          = use_body ?? true %}
 
 {# Common blocks : Declared this way to use `is not empty` and prevent the `has already been rendered` form field twig error #}
 {% set header %}{% block box_header %}{% endblock %}{% endset %}
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -61,9 +61,9 @@
     {# Body #}
     {% block box_body_before %}{% endblock %}
     {% if _hasBody %}
-    <div class="{% block box_type_class %}card-body {% endblock %}{% block box_body_class %}{% endblock %} {% if _fullsize %}p-0{% endif %} {% if _collapsible %} {{ _collapsible_class ~ (_collapsed ? ' collapse' : ' show') }} {% endif %}">
-        {{ body|raw }}
-    </div>
+        <div class="{% block box_body_classes %}card-body {% block box_body_class %}{% endblock %}{% endblock %} {% if _fullsize %}p-0{% endif %} {% if _collapsible %} {{ _collapsible_class ~ (_collapsed ? ' collapse' : ' show') }} {% endif %}">
+            {{ body|raw }}
+        </div>
     {% endif %}
     {% block box_body_after %}{% endblock %}
 
--- a/templates/embeds/card.html.twig
+++ b/templates/embeds/card.html.twig
@@ -8,7 +8,7 @@
 {% set _collapsible         = collapsible ?? _collapsed %}
 {% set _footer_collapsible  = footer_collapsible ?? true %}
 {% set _use_footer          = use_footer ?? false %}
-{% set _force_body          = force_body ?? false %}
+{% set _force_body          = force_body ?? true %}
 
 {# Common blocks : Declared this way to use `is not empty` and prevent the `has already been rendered` form field twig error #}
 {% set header %}{% block box_header %}{% endblock %}{% endset %}


kevinpapst:
Bildschirm­foto 2023-03-31 um 22 32 16

Can you try that:
```
        {% embed '@theme/embeds/card.html.twig' %}
            {% from "macros/widgets.html.twig" import work_times_result %}
            {% block box_type_class %}card-table table-responsive {% endblock %}
            {% block box_body %}
                <table class="table table-vcenter table-hover dataTable">
```

cavasinf:
Yeah it works!
But why do a simple div `.card-table` as parent won't..

I have:
1. A `_list.html.twig` partial file that is composed of that table:
```twig
<div class="table-responsive">
    <table class="table table-vcenter">
        <tbody>
        {# ... #}
        </tbody>
    </table>
</div>
```

2. An `index.html.twig` that call that partial inside a `.card`:
```twig
{% embed '@Tabler/embeds/card.html.twig' with {fullsize: true} %}
    {% block box_type_class %}card-table{% endblock %}
    {% block box_body %}
        {% include 'produit/_partials/_list.html.twig' with {produits : paginator.results} %}
    {% endblock %}
{% endembed %}
```

3. A modal that includes directly the `_list.html.twig` inside the modal body (not a card).
```twig
<div class="modal" tabindex="-1">
  <div class="modal-dialog">
    <div class="modal-content">
      <div class="modal-header">
        {# ... #}
      </div>
      <div class="modal-body">
        {% include 'produit/_partials/_list.html.twig' with {produits : paginator.results} %}
      </div>
      <div class="modal-footer">
        {# ... #}
      </div>
    </div>
  </div>
</div>
```

(3) I don't want to have the `.card-table` class when using the modal.
(1) Adding that class inside the `_list` template sounds wrong.
(1) And conditionally add the class inside the `_list` sounds lazy too.

It is maybe a Tabler issue with how that class behavior works 🤔

templates/embeds/card.html.twig
```diff
@@ -8,6 +8,7 @@
 {% set _collapsible         = collapsible ?? _collapsed %}
 {% set _footer_collapsible  = footer_collapsible ?? true %}
 {% set _use_footer          = use_footer ?? false %}
+{% set _force_body          = force_body ?? false %}
```
cavasinf, line 11:
```suggestion
{% set _force_body          = force_body ?? true %}
```

Force by default to avoid BC BREAK

templates/embeds/card.html.twig
```diff
@@ -8,6 +8,7 @@
 {% set _collapsible         = collapsible ?? _collapsed %}
 {% set _footer_collapsible  = footer_collapsible ?? true %}
 {% set _use_footer          = use_footer ?? false %}
+{% set _force_body          = force_body ?? true %}
```
cavasinf, line 11:
```suggestion
{% set _use_body          = use_body ?? true %}
```

templates/embeds/card.html.twig
```diff
@@ -20,7 +21,7 @@
 {% set _hasTitle            = title is not empty %}
 {% set _hasTools            = tools is not empty %}
 {% set _hasHeader           = header is not empty %}
-{% set _hasBody             = body is not empty %}
+{% set _hasBody             = body is not empty or _force_body %}
```
cavasinf, line 24:
```suggestion
{% set _hasBody             = _use_body or body is not empty %}
```

cavasinf (commented):
Follows footer condition

templates/embeds/card.html.twig
```diff
@@ -59,9 +60,11 @@
 
     {# Body #}
     {% block box_body_before %}{% endblock %}
+    {% if _hasBody %}
```
kevinpapst, line 63: What about `{% block body_outer %}` instead? Feels more natural to me in an embed, instead of having tons of magic variables.

cavasinf:
> comes directly from preview.tabler.io.

Yeah, I've also searched for how they intend to use it.
Looking at the code, they have no real "way" to use it:
- Can be on `.table`
- Can be on `.table` when using `.table-responsive`
- Can be directly on `.table-responsive`

See here: https://github.com/tabler/tabler/search?q=card-table&type=code

cavasinf:
> If you add that extra block around, you don't have to use workarounds with body_before or body_after. What do you think?

No I don't think so..

Once again, I'm following Tabler of "How you should do it".
I've made a `card-paginator` template that allows me to:
1. _quick filter_ 
2. _number of entries_ on a page 
3. _paginate_ a table card element

That card extends "TablerBundle card" and uses thoses `xxx_before` and/or `xxx_after` to follow Tabler ways:

image

[See here](https://preview.tabler.io/tables.html) "invoices" table

ATE, this is why I've made this PR 😄 to be able to replicate exactly that card template.

kevinpapst:
That is IMO stupid from Tabler side.

What if you:
- move the search part inside `block box_body_before` with the classes `border-bottom p-3`
- use the new `block box_type_class` with the content `card-body table-responsive` 
- and then move the main table into `block body` 

Wouldn't that work?

cavasinf:
> That is IMO stupid from Tabler side.

😆 

> Wouldn't that work?

Maybe, but I'm brainlessly following what they say "how to do".
If the doc (there is no doc here 😉) says to do it that way, I'll do it that way, "stupid" or not 😆. 
And I think other developers will too, especially if they are not familiar with "front things".

To add more complexity, Tabler really likes to put `+` condition inside there selectors.
- Breadcrumb
- Buttons
- Button Group
- List group
- Double Card Body
- Card button
- Nav Tabs
- And more..

This means that if we don't stick to how they intend it, and move a class or wrap it with another div, 
one/several whole selectors are out.
(I've done this to a custom component, and it's a lot of pain to keep alive when updating the template...)

cavasinf:
Side note: I also use card with multiple bodies. And Tabler is okay with that.
https://github.com/tabler/tabler/blob/dev/dist/css/tabler.css#L18916-L18918

image

cavasinf:
> Can you please give it a try?

Sure!
And it works:
image

> That is IMO stupid from Tabler side.

I join you on that point, it more natural to me to do it that way..

I agree, it is more natural for me to do it that way.

ATE there is:
- `box_header` block content
- box_header_after' block content
- No box body:
   1. Can be removed using `box_type_class
   2. (With this PR) No `box_body` block and use `block_body_before` OR `block_body_after` block.
- Content of the `box_footer` block

cavasinf:
TBH, I really don't like `box_type_class`:
1. I'll never look to that name to remove a `.box-body` class.
2. `boxtype` param is already used to define the box status (which I don't like either)

I'd rather have a `box_body_classes' that wraps each `box_body' class.
Or (this PR) not printing the body at all.

I can give other examples that I use today.
### 1. If I want to do a ribbon without a card body:
```html
<div class="ribbon ribbon-top bg-yellow">
   <!-- Content -->
</div>
```
I'll not replace the `box_type_class` to add `ribbon` classes.
My though will be to not print the body and place my custom ribbon after the header or before the body.

```html
<div class="card">
  <div class="card-header">
    <h3 class="card-title">Card title</h3>
  </div>
  <div class="ribbon ribbon-top bg-yellow">
    <svg></svg>
  </div>
  <!-- NO body-->
</div>
```

image


### 2. Same case for a card with a progress bar `.card-progress`
````html
<div class="progress progress-sm card-progress">
  <div class="progress-bar" style="width: 38%" role="progressbar" aria-valuenow="38" aria-valuemin="0" aria-valuemax="100" aria-label="38% Complete">
    <span class="visually-hidden">38% Complete</span>
  </div>
</div>
```
![image](https://user-images.githubusercontent.com/25293190/230310029-55678c2b-41b6-4585-a761-d6b178c90a13.png)

kevinpapst:
The name is not ideal but `box_body_class` was already used and moving the `card_body` class into `box_body_class` would have broken all of our apps. That why I went with `box_type_class`. Not great naming probably... but I only accept criticism, if you come up with a better alternative 😁 and we can still change it, I doubt that it is widely used yet.

That's why I proposed a block around the body div, so you can easily empty/remove it. But with the current setup, you should be able to use the box_type_class to achieve both your progress:

```twig
{% block box_type_class %}progress progress-sm card-progress{% endblock %}
{% block box_body %}
  <div class="progress-bar" style="width: 38%" role="progressbar" aria-valuenow="38" aria-valuemin="0" aria-valuemax="100" aria-label="38% Complete">
    <span class="visually-hidden">38% Complete</span>
  </div>
{% endblock %}
```
and ribbon example:
```twig
{% block box_type_class %}ribbon ribbon-top bg-yellow{% endblock %}
{% block box_body %}
    <svg></svg>
{% endblock %}
```

And with all the learnings from the card monster, we can create a card-v2 embed later that cleans up that "creative" mess.

cavasinf:
> I'd rather have a `box_body_classes` that wraps each box_body class.

you don't like that one?

kevinpapst:
Ah, I missed that one. Your posts are always so long 🙈 

You mean like that:
``` 
<div class="{% block box_body_classes %}card-body {% block box_body_class %}{% endblock %}{% endblock %}
```
?
Looks good. I should have come up with that. Yes, lets do it and add the block around the entire body, so it can be removed.

cavasinf:
> Your posts are always so long 🙈

Yeah I don't know why 😆 
French people like to talk 🤔 Maybe..

------

`<div class="{% block box_body_classes %}card-body {% block box_body_class %}{% endblock %}{% endblock %}`

Looks good, but here's why I did that PR.
What do we do for others classes ?
- fullsize param
- collapisble param

```twig
{% if _fullsize %}p-0{% endif %} {% if _collapsible %} {{ _collapsible_class ~ (_collapsed ? ' collapse' : ' show') }} {% endif %}"
```

Do I need the fullsize condition if I want to do a ribbon?
Do I need the collapse condition if I want to do a ribbon?

So why using box_body block at all if i want to remove everything?

I know I'm a pain in the ass, but I'd rather we all agree than skipping things. (sorry 😝)

kevinpapst:
And here is why I proposed the block instead of a variable: you have more control for yet unknown use-cases. I bet at some point one of us wants to use the body and put something entirely different in there.

Anyway ... looks good to me now after the discussion! And I will release it directly after merging, as I need to fix my uses of the box_body_type class.

templates/embeds/card.html.twig
```diff
@@ -59,9 +60,11 @@
 
     {# Body #}
     {% block box_body_before %}{% endblock %}
-    <div class="{% block box_type_class %}card-body {% endblock %}{% block box_body_class %}{% endblock %} {% if _fullsize %}p-0{% endif %} {% if _collapsible %} {{ _collapsible_class ~ (_collapsed ? ' collapse' : ' show') }} {% endif %}">
-        {{ body|raw }}
-    </div>
+    {% if _hasBody %}
```
kevinpapst, line 63: stop, one more thing: this is a BC break.
Now the body will be removed if it is empty as well. Why did we not use the `_hasBody` variable before? 
Did we not add that check before on purpose? 

I just checked my app and it will likely not hit me, but I could imagine, that people might have used the embed without defined body and then use javascript to write into it, e.g. for dynamic modal content?
You think that might become a problem?

cavasinf: That was in my first commit and PR version
> Can be BC break, if some JS is made to fill that card-body 🤔

ATE I've removed this mention because `use_body` is set to `true` by default.
So it is the same behavior as today. (no BC break then)

In MY case I have to set use_body to `false` to not print it.

In the V2 card we can remove this 👍

kevinpapst: 1. What is [ATE](https://www.abbreviationfinder.org/acronyms/ate.html)?
2. What a long story. Sorry! I didn't get any of it in the first place I guess. Probably I am working too much on my own 😁 Actually using force_body from your initial commit sounds more reasonable to me for the long term. If the body block is not defined, than the container will not be rendered, unless you force it to.

I am fine with both solutions, actually that would be a minor BC, which I would be okay with. 

You are the maintainer as well, so what do you think is better? 
use_body = true || force_body = false

kevinpapst (approved):

PR #156, opened 2023-04-19
[Alert] Add `id` and `attr` option

cavasinf:
## Description
Adds `id` and `attr` option

## Types of changes

## Checklist

### DIFF
--- a/templates/components/alert.html.twig
+++ b/templates/components/alert.html.twig
@@ -19,7 +19,7 @@
     {% set _background = _title is empty or _description is empty or _important %}
 
     <div
-            {% if _id -%}id="{{ _id }}"{%- endif %}
+            {%- if _id %}id="{{ _id }}"{% endif -%}
             class="alert alert-{{ _type }} {% if _dismissible %}alert-dismissible{% endif %} {% if _background %}alert-important{% endif %} {{ _class }}"
             role="alert"
             {{- macros.attr_to_html(_attr) -}}
--- a/templates/components/alert.html.twig
+++ b/templates/components/alert.html.twig
@@ -1,5 +1,5 @@
 {% macro alert(options) %}
-    {% import '@Tabler/includes/utils.html.twig' as macros %}
+    {% import '@Tabler/includes/utils.html.twig' as utils %}
 
     {# Content #}
     {% set _title = options.title ?? null %}
@@ -16,13 +16,16 @@
     {% set _class = options.class ?? '' %}
     {% set _raw = (options.raw ?? false) is same as (true) %}
 
-    {% set _background = _title is empty or _description is empty or _important %}
+    {% if (_title is empty or _description is empty) and options.important is not defined %}
+        {% deprecated "Using important alert without setting the 'important' option is deprecated, set option to true instead." %}
+        {% set _important = true %}
+    {% endif %}
 
     <div
-            {%- if _id %}id="{{ _id }}"{% endif -%}
-            class="alert alert-{{ _type }} {% if _dismissible %}alert-dismissible{% endif %} {% if _background %}alert-important{% endif %} {{ _class }}"
+            {% if _id %}id="{{ _id }}"{% endif %}
+            class="alert alert-{{ _type }} {% if _dismissible %}alert-dismissible{% endif %} {% if _important %}alert-important{% endif %} {{ _class }}"
             role="alert"
-            {{- macros.attr_to_html(_attr) -}}
+            {{ utils.attr_to_html(_attr) }}
     >
         <div class="d-flex">
             {% if _icon %}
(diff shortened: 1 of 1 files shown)


kevinpapst:
Yes, remove the deprecated stuff. 

I know you like the better readability of multiline HTML. 
But I have several test cases where I check the generated HTML for special cases (sometimes that is the only way to test a certain code piece. Those tests always break with HTML structure changes. Can you use {%- and -}} to prevent line breaks? 

What just came to my mind while reading your changes: we could merge `role` and `id` into the `_attr` array and let the macro create all HTML attributes. The macro could prevent line breaks, my tests are happy about the HTML and you have the better readability.

templates/components/alert.html.twig
```diff
@@ -1,43 +1,28 @@
...
+    {% set _id = options.id ?? null %}
+    {% set _attr = options.attr ?? {} %}
+    {% set _class = options.class ?? '' %}
+    {% set _raw = (options.raw ?? false) is same as (true) %}
 
     {% set _background = _title is empty or _description is empty or _important %}
 
     <div
-            {% if _id %}id="{{ _id }}"{% endif %}
+            {% if _id -%}id="{{ _id }}"{%- endif %}
```
kevinpapst, line 22: Isn't it the opposite way `{%- if _id %}id="{{ _id }}"{% endif -%}`?

cavasinf:
Talking about things to change.
I don't think that:
> {% set _background = _title is empty or _description is empty or _important %}

Has its place in the bundle.
Maybe it is related to your logic or the old AdminLTE logic.
But I think the `alert-important` class should be written when the `important` option is set to `true`.

templates/components/alert.html.twig
cavasinf: TRUE!

```suggestion
            {%- if _id %}id="{{ _id }}"{% endif -%}
```

templates/components/alert.html.twig
```diff
@@ -1,32 +1,29 @@
...
+    {% set _id = options.id ?? null %}
+    {% set _attr = options.attr ?? {} %}
+    {% set _class = options.class ?? '' %}
+    {% set _raw = (options.raw ?? false) is same as (true) %}
 
     {% set _background = _title is empty or _description is empty or _important %}
 
-    <div class="alert alert-{{ _type }}{% if _dismissible %} alert-dismissible{% endif %}{% if _background %} alert-important{% endif %}{{ _class }}" role="alert">
+    <div
+            {%- if _id %}id="{{ _id }}"{% endif -%}
```
kevinpapst, line 22: Ah, I am sorry, is there now a whitespace missing? 
```suggestion
            {%- if _id %}id="{{ _id }}" {% endif -%}
```

kevinpapst:
Where do I find `{% set _background = _title is empty or _description is empty or _important %}` that?

I had recently troubles showing an alert notification in a certain way. Maybe that was the reason...

templates/components/alert.html.twig
cavasinf: With a tilde `~` looks like to work!
```suggestion
            {%- if _id %}id="{{ _id }}"{% endif ~%}
```

kevinpapst, line 22: That should do the trick
```suggestion
    <div {%- if _id %}id="{{ _id }}" {% endif -%}
```

cavasinf:
> Where do I find `{% set _background = _title is empty or _description is empty or _important %}` that?
> 
> I had recently troubles showing an alert notification in a certain way. Maybe that was the reason...

There: https://github.com/kevinpapst/TablerBundle/blob/4148b2404d70ff46dcaa0ed18c1312ce3c438ab1/templates/components/alert.html.twig#L19

templates/components/alert.html.twig
cavasinf: Produces that:
```
<divclass="alert alert-danger alert-dismissible alert-important "
        role="alert"
>
```

kevinpapst: Ah damn, I am stressed, sorry. I forgot to remove - 
```suggestion
    <div {% if _id %}id="{{ _id }}" {% endif -%}
```

cavasinf: Wrapping all the code by `{% filter spaceless %}` doesn't change the attributes spaces of the `div` parent.

With:
```twig 
{{ alert({
    icon: 'help',
    title: 'title',
    description: 'description',
    dismissible: true,
    important: true,
}) }}
```

Produces:
```html
<div
                        class="alert alert-danger alert-dismissible alert-important "
            role="alert"
            
    ><div class="d-flex"><div class="alert-icon"><i class="icon far fa-question-circle"></i></div><div><h4 class="mb-1">title</h4><p class="m-0">description</p></div></div><a class="btn-close" data-bs-dismiss="alert" aria-label="close"></a></div>
```

Is it even possible to remove space inside HTML tag? 🤔

cavasinf: Reverting that due to breaking code

cavasinf:
Any update on that @kevinpapst?

kevinpapst (approved):
Tested, works 👍 merge 🚀 

Thanks!

PR #162, opened 2023-05-30
allow collapse on complete card header

kevinpapst:
## Description

Customers mentioned, that it would be much more comfortable, if they could click the entire card header.

So not only the arrow icon, but also the entire title and whitespace in between:
Bildschirmfoto 2023-05-30 um 17 06 40

I have to agree, they are not wrong.

WDYT @cavasinf ?

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] I updated the documentation (see [here](https://github.com/kevinpapst/TablerBundle/tree/master/docs))
- [x] I agree that this code will be published under the [MIT license](https://github.com/kevinpapst/TablerBundle/blob/master/LICENSE)

PR #165, opened 2023-05-31
added combined card-accordion component

kevinpapst:
## Description

Replaces #162 

## Types of changes

## Checklist

### DIFF
--- a/templates/embeds/collapsible.html.twig
+++ b/templates/embeds/collapsible.html.twig
@@ -1,27 +1,19 @@
-{% extends '@Tabler/embeds/card.html.twig' %}
-{% set collapsed = false %}
 {% set fullsize = true %}
+{% extends '@Tabler/embeds/card.html.twig' %}
+{% from '@Tabler/components/accordion.html.twig' import accordion %}
+
+{% block box_class %}border-0{% endblock %}
 {% block box_title %}{% endblock %}
+
+{% set item = {
+    title: block('title') ?? '',
+    body: block('body') ?? '',
+    bodyClasses: '',
+    options: { open: open ?? false, bodyExtraClass: ((border ?? true) ? 'border-top pt-3' : '') }
+} | merge(item ?? {}) %}
+
+{% set options = {flush: flush ?? false} | merge(options ?? {}) %}
+
 {% block box_body %}
-    {% set _id = id ?? tabler_unique_id('collapsible_') %}
-    {% set _open = open ?? false %}
-    {% set _id_collapse = _id ~ '_item' %}
-
-    <div id="{{ _id }}" class="accordion accordion-flush{% block class %} {% endblock %}">
-        <div class="accordion-item {% block item_class %} {% endblock %}">
-            <div class="accordion-header" id="{{ _id }}_heading">
-                <button class="accordion-button {% if not _open %}collapsed{% endif %}"
-                        type="button" data-bs-toggle="collapse" data-bs-target="#{{ _id_collapse }}"
-                        aria-expanded="true">
-                    {% block title %}{% endblock %}
-                </button>
-            </div>
-
-            <div id="{{ _id_collapse }}" class="accordion-collapse collapse {% if _open %}show{% endif %}">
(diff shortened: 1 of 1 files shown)


templates/embeds/collapsible.html.twig
```diff
@@ -0,0 +1,27 @@
+{% extends '@Tabler/embeds/card.html.twig' %}
+{% set collapsed = false %}
```
cavasinf, line 2: Is this still needed?

templates/embeds/collapsible.html.twig
```diff
@@ -0,0 +1,27 @@
...
+{% set fullsize = true %}
```
cavasinf, line 3: Parameter under `extends` works as untended?
I always put them above the extends , even more if the extend use it.

templates/embeds/collapsible.html.twig
```diff
@@ -0,0 +1,27 @@
...
+{% block box_title %}{% endblock %}
+{% block box_body %}
+    {% set _id = id ?? tabler_unique_id('collapsible_') %}
+    {% set _open = open ?? false %}
+    {% set _id_collapse = _id ~ '_item' %}
```
cavasinf, line 8: Can we put that on top?
(part of me want that 😄)

templates/embeds/collapsible.html.twig
```diff
@@ -0,0 +1,27 @@
...
+                </button>
+            </div>
+
+            <div id="{{ _id_collapse }}" class="accordion-collapse collapse {% if _open %}show{% endif %}">
+                <div class="accordion-body {% block body_class %}pt-0{% endblock %}">
+                    {% block body %}{% endblock %}
+                </div>
+            </div>
+        </div>
+    </div>
```
cavasinf, line 26: I understand that the `accordion` component is scary and not easy to manipulate/use.
But isn't it more "Maintainers" friendly to use it?
The less duplicate code we have to maintain, the happier we are?

cavasinf:
Still using blocks `title` and `body`.
I'll go like that:
```twig
{% set fullsize = true %}
{% extends '@Tabler/embeds/card.html.twig' %}
{% from '@Tabler/components/accordion.html.twig' import accordion %}

{% block box_class %}border-0{% endblock %}
{% block box_title %}{% endblock %}

{% set _accordionItem = {
    title: block('title') ?? '',
    body: block('body') ?? '',
    bodyClasses: '',
    options: {
        open: false,
        bodyExtraClass: 'border-top',
    }
} | merge(accordionItem ?? {}) %}

{% set _accordionOption = {} | merge(accordionOption ?? {}) %}

{% block box_body %}
    {{ accordion(
        [_accordionItem],
        _accordionOption
    ) }}
{% endblock %}
```

Used like that:
````twig
{% set item = {
    options : {
        open : true
    }
} %}
{% embed '@theme/embeds/collapsible.html.twig' with {accordionItem  : item} %}
    {% block title %}File permissions{% endblock %}
    {% block body %}
        .....
    {% endblock %}
{% endembed %}
```

WDYT @kevinpapst?

kevinpapst:
I'll give this a try later today ... full day of meetings.
Looks like a clever solution indeed, better than mine 😁

templates/embeds/collapsible.html.twig
kevinpapst: To me an extends belong to the top of the file, to understand the structure.
But yes, it works even below. If you look at the compiled class for this template, you'll understand why.

templates/embeds/collapsible.html.twig
kevinpapst: Yes, The should always be collapsed, and then you define through the accordionItem `open` flag if the content is visible or not.

kevinpapst:
Small adjustments applied, to simplify passing options.

```
    {% embed '@theme/embeds/collapsible.html.twig' with { boxtype: 'danger', border: false, open: true, flush: true } %}
        {% block title %}Foo{% endblock %}
        {% block body %}Bar{% endblock %}
```

Can be reviewed @cavasinf

PR #170, opened 2023-09-23
Version 1.0

kevinpapst:
## Description

- Symfony 6 only
- PHP 8 only
- Phpstan level 9
- Fixes deprecation with SF 6.3 about missing return types

Besides the obvious BC breaks due to added return types, there is one BC break that needs attention.

The `UserInterface` whose method changed from `getIdentifier()` to `getUserIdentifier()`.
As latter method is part of the Symfony User Interface, this shouldn't be an issue, as we target SF 6 only.

## Types of changes

## Checklist

### DIFF
--- a/src/Model/NotificationModel.php
+++ b/src/Model/NotificationModel.php
@@ -13,16 +13,10 @@ use KevinPapst\TablerBundle\Helper\Constants;
 
 class NotificationModel implements NotificationInterface
 {
-    private $id;
-    private $message;
-    private $type;
-    private $url;
+    private ?string $url = null;
 
-    public function __construct(string $id, string $message, string $type = Constants::TYPE_INFO)
+    public function __construct(private string $id, private string $message, private string $type = Constants::TYPE_INFO)
     {
-        $this->id = $id;
-        $this->message = $message;
-        $this->type = $type;
     }
 
     public function getIdentifier(): string
--- a/src/Model/UserModel.php
+++ b/src/Model/UserModel.php
@@ -11,27 +11,11 @@ namespace KevinPapst\TablerBundle\Model;
 
 final class UserModel implements UserInterface
 {
-    /**
-     * @var string
-     */
-    private $id;
-    /**
-     * @var string
-     */
-    private $name;
-    /**
-     * @var string|null
-     */
-    private $title;
-    /**
-     * @var string|null
-     */
-    private $avatar;
+    private ?string $title = null;
+    private ?string $avatar = null;
 
-    public function __construct(string $id, string $name)
+    public function __construct(private string $id, private string $name)
     {
-        $this->id = $id;
-        $this->name = $name;
     }
 
     public function getId(): string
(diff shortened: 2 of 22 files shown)


tacman:
This is great!   Can you also add Symfony 7, which will be out in 2 months?

My branch of the demo works with Symfony 6.4 and I've been removing deprecation warnings.

src/Model/NotificationModel.php
```diff
@@ -13,16 +13,10 @@
...
 class NotificationModel implements NotificationInterface
 {
-    private $id;
-    private $message;
-    private $type;
-    private $url;
+    private ?string $url = null;
 
-    public function __construct(string $id, string $message, string $type = Constants::TYPE_INFO)
+    public function __construct(private string $id, private string $message, private string $type = Constants::TYPE_INFO)
```
cavasinf, line 18: I'm not a big fan of putting class members INSIDE the constructor.
We have encountered a lot of confusing behavior logic when using it that way (in our company).
Even more in a single line! It will be a pain when resolving conflicts on merge! (trust me 😭)

No complaint when using autowire though, just on models like this one.

I'll go:
```php
public function __construct(
    private string $id, 
    private string $message, 
    private string $type = Constants::TYPE_INFO
) {
}
```

Or back to OLD way
```php
private string $id, 
private string $message, 
private string $type = Constants::TYPE_INFO
private ?string $url = null;

public function __construct(
    string $id, 
    string $message, 
    string $type = Constants::TYPE_INFO
) {
        $this->id = $id;
        $this->message = $message;
        $this->type = $type;
}
```

src/Model/UserModel.php
```diff
@@ -11,27 +11,11 @@
...
-    private $title;
-    /**
-     * @var string|null
-     */
-    private $avatar;
+    private ?string $title = null;
+    private ?string $avatar = null;
 
-    public function __construct(string $id, string $name)
+    public function __construct(private string $id, private string $name)
```
cavasinf, line 17: Same here

src/Model/NotificationModel.php
cavasinf: If you are using `rector` to do this PR, we have added the `Rector\Php80\Rector\Class_\ClassPropertyAssignToConstructorPromotionRector` to the `skip` config. And do this upgrade only where needed.

cavasinf:
What about [those issues](https://github.com/kevinpapst/TablerBundle/milestone/1) @kevinpapst?

kevinpapst:
My changes are BC breaks, that's why we need to bump to 1.0.
Also I am not a big fan of the "I am waiting for years until 1.0" release tactic. 
It is time to signal that the bundle is stable.
And there is always a time for new major release if we do BC breaks in the future (e.g. remove |trans filter).

Lets move the tickets to a new milestone and close the linked one.
SF 6.3 is out and there are deprecations that need to be fixed now.
SF7 is coming soon, we should be prepared.

I agree, we can always release a 1.1 to support SF7 in two month.

Regarding your code comments: these files were NEVER changed since I committed them 2 years ago.
I am working mostly by myself since years, so you have better insights in merge problems. Larger teams on active codebases obviously have different needs than projects like this one 😄  
But I do not see a big risk here, due to the rare commits in the last months 🤷 which is actually a good sign of stability IMO.

PR #176, opened 2023-11-03
[UserNotification] Add more configuration options

cavasinf:
## Description
The current NotificationEvent is not complete enough to allow devs to do what they want.
Updated it as Tabler `1.0.0-beta20` style.

user_notifications

```php
<?php

namespace App\EventSubscriber;

use KevinPapst\TablerBundle\Event\NotificationEvent;
use KevinPapst\TablerBundle\Helper\Constants;
use KevinPapst\TablerBundle\Model\NotificationModel;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;

class TablerNotificationSubscriber implements EventSubscriberInterface
{
    public static function getSubscribedEvents(): array
    {
        return [
            NotificationEvent::class => ['onSetupNotification', 100],
        ];
    }

    public function onSetupNotification(NotificationEvent $event): void
    {
        $event->title          = 'My custom title';
        $event->emptyTitle     = 'No Item custom title';
        $event->badgeColor     = 'green';
        $event->showBadgeTotal = true;
        $event->maxDisplay     = 7;

        $activeNotification = new NotificationModel('active', 'My active Message', Constants::TYPE_WARNING);
        $activeNotification->setActive(true);
        $event->addNotification($activeNotification);

        // This one will not be showed as it is invalid
        $invalidNotification = new NotificationModel('invalid');
        $event->addNotification($invalidNotification);

        $defaultNotification = new NotificationModel('default', 'My default Message');
        $event->addNotification($defaultNotification);

        $disabledNotification = new NotificationModel('disabled', 'My disabled Message', null);
        $disabledNotification->setDisabled(true);
        $disabledNotification->setBadgeAnimated(false);
        $event->addNotification($disabledNotification);

        $customizeNotification = new NotificationModel('customize', 'Link to Google', Constants::TYPE_SUCCESS);
        $customizeNotification->setBadgeAnimated(false);
        $customizeNotification->setUrl('https://www.google.com');
        $event->addNotification($customizeNotification);

        $htmlNotification = new NotificationModel('html');
        $htmlNotification->setHtml('
            <div class="list-group-item">
                <div class="row align-items-center">
                  <div class="col-auto"><span class="status-dot status-dot-animated bg-red d-block"></span></div>
                  <div class="col text-truncate">
                    <a href="#" class="text-body d-block">Example 1</a>
                    <div class="d-block text-muted text-truncate mt-n1">
                      Change deprecated html tags to text decoration classes (#29604)
                    </div>
                  </div>
                  <div class="col-auto">
                    <a href="#" class="list-group-item-actions">
                      <svg xmlns="http://www.w3.org/2000/svg" class="icon text-muted" width="24" height="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round"><path stroke="none" d="M0 0h24v24H0z" fill="none"></path><path d="M12 17.75l-6.172 3.245l1.179 -6.873l-5 -4.867l6.9 -1l3.086 -6.253l3.086 6.253l6.9 1l-5 4.867l1.179 6.873z"></path></svg>
                    </a>
                  </div>
                </div>
            </div>
        ');
        $event->addNotification($htmlNotification);

        $badgeNotification = new NotificationModel('badge');
        $badgeNotification->setHtml('
            <li class="list-group-item d-flex justify-content-between align-items-center">
                Notification with badge
                <span class="badge badge-primary badge-pill">14</span>
            </li>
        ');
        $event->addNotification($badgeNotification);

        $flexBoxNotification = new NotificationModel('flexbox');
        $flexBoxNotification->setHtml('
            <a href="#" class="list-group-item list-group-item-action flex-column align-items-start">
                <div class="d-flex w-100 justify-content-between">
                  <h5 class="mb-1">List group item heading</h5>
                  <small>3 days ago</small>
                </div>
                <p class="mb-1">Donec id elit non mi porta gravida at eget metus. Maecenas sed diam eget risus varius blandit.</p>
                <small>Donec id elit non mi porta.</small>
            </a>
        ');
        $event->addNotification($flexBoxNotification);

        $moreThanMaxNotification =
            new NotificationModel('max', 'Will not be displayed as max notification is set to 7');
        $event->addNotification($moreThanMaxNotification);

        $extraNotification = new NotificationModel('extra', 'One more not displayed');
        $event->addNotification($extraNotification);
    }
}
```

### When there is an Event without any Notification:
image

```php
public function onSetupNotification(NotificationEvent $event): void
    {
        $event->title          = 'My custom title';
        $event->emptyTitle     = 'No Item custom title';
        $event->badgeColor     = 'green';
        $event->showBadgeTotal = true;
        $event->maxDisplay     = 7;
        
        // simulate no notification
        return;
}
```

### Without any subscriber:
image

### How it is today

|  With notifications | Without notification |  Without subscriber |
|---|---|---|
| image| image| image|

## Types of changes

## Checklist


cavasinf:
Hey ✋, it's been a while!

src/Event/NotificationEvent.php
```diff
@@ -13,22 +13,37 @@
 
 class NotificationEvent extends ThemeEvent
 {
-    /**
-     * @var array<NotificationInterface>
-     */
+    public ?string $title = null;
```
kevinpapst, line 16: no public variables. only setter and getter

src/Model/NotificationModel.php
```diff
@@ -13,39 +13,133 @@
...
+    // Custom HTML
+    private ?string $html = null;
+
+    public function isValid(): bool
     {
+        return $this->message || $this->html;
+    }
+
+    public function __construct(
+        string $id,
```
kevinpapst, line 36: Constructor property promotion! This is PHP 8... 😁 but seriously, not that old-school redudancy please

src/Model/NotificationModel.php
```diff
@@ -13,39 +13,133 @@
...
-        return $this->type;
+        return $this->message;
     }
 
-    public function setMessage(string $message): void
+    public function setMessage(?string $message): self
     {
         $this->message = $message;
+
+        return $this;
```
kevinpapst, line 59: 👎 for fluent interfaces

src/Event/NotificationEvent.php
```diff
@@ -13,22 +13,37 @@
...
-     * @return NotificationInterface[]
+     * @return array<int,NotificationInterface>
      */
     public function getNotifications(?int $max = 10): array
     {
-        if (null !== $max) {
-            return \array_slice($this->notifications, 0, $max);
+        $notifications = array_filter(
+            $this->notifications,
+            fn(NotificationInterface $notification) => $notification->isValid()
```
kevinpapst, line 37: never used that syntax. which php version allows that ?

kevinpapst (changes requested):
Welcome back 😄 👋 

I like the proposed changes! Just one thing that should be configurable => display of the menu always vs. only if messages exist.

The rest is feedback on code style. I know that we kinda disagree on certain code styles, but I hope you can live with my wishes. Not your main project, so you don't have to see it every day 😁

templates/includes/navbar_notifications.html.twig
```diff
@@ -1,22 +1,68 @@
...
-                    </a>
-                {% endfor %}
-            </div>
-        </div>
-    </div>
-{% endif %}
+{# @var notificationEvent \KevinPapst\TablerBundle\Event\NotificationEvent #}
+{% set notificationEvent = tabler_notifications() %}
+
+{% if notificationEvent %}
```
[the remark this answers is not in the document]
cavasinf: It does return `null` when no listener:
https://github.com/kevinpapst/TablerBundle/blob/0fc642e67e77c5ea555395f27d4427969e3443c2/src/Twig/RuntimeExtension.php#L85-L95

I like the `setShowIfEmpty`. I'll add it 👍

src/Model/NotificationInterface.php
```diff
@@ -1,21 +1,33 @@
...
+/*
+ * This file is part of the Tabler bundle, created by Kevin Papst (www.kevinpapst.de).
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace KevinPapst\TablerBundle\Model;
+
+interface NotificationInterface
```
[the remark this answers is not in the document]
cavasinf: I'm not sure I get it.
What is BC Breaks? The changes of return type?

src/Model/NotificationModel.php
cavasinf: Yeah I know 😞
I really don't like to use those when we have properties IN class + constructor.
Look a bit messy for me.
But I use it a lot with dependency injection!

Anyway, I'll change that.

src/Model/NotificationModel.php
cavasinf: OK, can you explain me why?
(interface is only for getter not setter)

src/Event/NotificationEvent.php
cavasinf: MB !!! 💀

src/Event/NotificationEvent.php
cavasinf: It is a "js" way to make function.
Called ["Arrow function" from PHP 7.4](https://www.php.net/manual/en/functions.arrow.php)

src/Model/NotificationInterface.php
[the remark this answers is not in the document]
kevinpapst: Literally any change on an interface is a BC break 🤷 Now that we are "1.x" we should limit these do a decent amount.  
The problem is: bumping composer to a new major requires every user to change composer files.
And many will not notice for a longer time that there is a new major available.

If we do not bump to a new major, we break semver and (could also break) codebases during composer update. 

What we could do to prevent that: 
- getMessage() stays as is and also returns the HTML
- type is always set, remove the ? from getType() return type
- add a second interface "NotificationNextGen" that contains the new methods, including a new one for `isHtml(): bool´

Now wrap all Notification objects in the NotificationEvent that are not `instanceof NotificationNextGen` in a NotificationModel (and trigger a deprecation). 
That should be enough as BC layer and would allow us to go forward without a new major version.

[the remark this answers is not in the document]
cavasinf: Can you take a look at https://github.com/kevinpapst/TablerBundle/pull/176/commits/bcff26e7c2c57124b465982dcba9719340c4475e please.

I think this made it messy.
Possible alternative, can we create a new model `NotificationHtmlModel` that can only contain HTML (string),
and remove all the notification logic (active, disabled, withBadge, badgeAnimated) from there?
In the end, `NotificationEvent->notifications` can handle `NotificationModel` AND `NotificationHtmlModel`.
And it avoids the BC break more:
- NotificationModel now only has new fields (active, disabled,..) without changing the olds one.
- NotificationHtmlModel was unknown before.

src/Model/NotificationModel.php
kevinpapst: Not agree on all he wrote down, but a lot of good points:
https://ocramius.github.io/blog/fluent-interfaces-are-evil/

But I don't think that we should do that, because the simple question is: what is the real benefit we win and what do we give up?

translations/TablerBundle.fr.xlf
```diff
@@ -146,6 +146,18 @@
...
+          <source>Show notifications</source>
+          <target>Afficher notifications</target>
+      </trans-unit>
+      <trans-unit id="F5Dfzd" resname="You have no notification">
+          <source>You have no notification</source>
+          <target>Vous avez aucune notification</target>
+      </trans-unit>
+      <trans-unit id="Pf48Zd" resname="And %count% more...">
+          <source>And %count% more...</source>
+          <target>Et %count% autres...</target>
```
kevinpapst, line 159: https://www.htmlsymbol.com/html-entities/horizontal-ellipsis.html

cavasinf: Never thought of using it.
Do people from weblate will do it too?

kevinpapst: I hope so 😁 Translators usually know these things, actually I learned it from one of them.
I just think it looks better than three singled dots

src/Model/NotificationModel.php
cavasinf: Yeah, I don't agree either on some point.
OK with the readability, and not enough experiences on my side with Proxy classes.
But the rest of it is kinda "meh". 
Maybe if he does that article with the current PHP version (not PHP 5.5 - 2013).

--------------

Anyway, if you don't want to, we'll not to.
The benefit are:
1. From 
  ```php
    $notification = new NotificationModel('1', '...');
    $notification->setMessage("Oops wrong site");
    $notification->setUrl("www.google.com");
    $notification->setActive(true);
    $notification->setBadgeAnimated(false);
  ```
2. To
  ```php
   $notification = (new NotificationModel('1', '...'))
        ->setMessage("Oops wrong site")
        ->setUrl("www.google.com")
        ->setActive(true)
        ->setBadgeAnimated(false)
    ;
  ```
That's it.

Cons:
At our level (not a framework), I don't know 🤷‍♂️

cavasinf:
Waiting for information in https://github.com/kevinpapst/TablerBundle/pull/176#discussion_r1383532681

cavasinf:
OK so:

> You could add a new NotificationInterface that extends the current one, holding all the new methods

Done

> Change the event (?) and accept both NotificationInterfaces to support both interfaces (maybe that is not necessar, as the new Interface extends the old one)

To be safe, I've added both interface in parameter.
If we want, we can add condition to trigger deprecated (commented today, see `addNotification()`)

> Change the template to support both interfaces

Instead of that, all the non defined properties from the old interface has default value from twig.
eg : `{% if notification.active ?? false %}active{% endif %}`

cavasinf:
At the end, we have two interfaces `NotificationV2Interface` and `NotificationInterface`.
If devs have custom model that:
1. `extends` from `NotificationModel`, there will be no compatibility issues if they do not have customized properties with the same name as this pull request. 
2. `implements` `NotificationInterface`, the default values from twig will be used.

BUT if they overwrite the `navbar_notifications.html.twig` from their project, errors will occur.

src/Model/NotificationV2Interface.php
```diff
@@ -0,0 +1,24 @@
...
+ * This file is part of the Tabler bundle, created by Kevin Papst (www.kevinpapst.de).
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace KevinPapst\TablerBundle\Model;
+
+// TODO : Implement these in NotificationInterface later
+interface NotificationV2Interface extends NotificationInterface
```
kevinpapst, line 13:
```suggestion
/**
 * TODO : Implement these in NotificationInterface later
 * @internal THIS WILL BE REMOVED WITH THE NEXT MAJOR RELEASE, NO BC PROMISE GIVEN
 * @deprecated
 */
interface NotificationV2Interface extends NotificationInterface
```

src/Event/NotificationEvent.php
```diff
@@ -9,35 +9,152 @@
 
 namespace KevinPapst\TablerBundle\Event;
 
+use KevinPapst\TablerBundle\Model\NotificationV2Interface;
 use KevinPapst\TablerBundle\Model\NotificationInterface;
 
-class NotificationEvent extends ThemeEvent
+class NotificationEvent extends ThemeEvent implements NotificationEventInterface
```
kevinpapst, line 15: NotificationEvent cannot be replaced with custom implementation, so the interface can be removed

templates/includes/navbar_notifications.html.twig
```diff
@@ -1,22 +1,79 @@
...
+                    <div class="list-group list-group-flush list-group-hoverable">
+                        {% for notification in notificationEvent.notifications %}
+                            {% if notification.html ?? false %}
+                                {{ notification.message | raw }}
+                            {% else %}
+                                <div
+                                    class="list-group-item {% if notification.active ?? false %}active{% endif %} {% if notification.disabled ?? false %}disabled{% endif %}"
+                                >
+                                    <div class="row align-items-center">
+                                        {% if notification.withBadge ?? true %}
```
kevinpapst, line 52: Does that work if it is an "old NotificationInterface" but not a "NotificationModel" ?

kevinpapst (changes requested):

src/Event/NotificationEvent.php
cavasinf:
```suggestion
class NotificationEvent extends ThemeEvent
```

templates/includes/navbar_notifications.html.twig
cavasinf: Yes every null-coalescing, fallback to the right operand if not exist / undefined / null.

> {# returns the value of foo if it is defined and not null, 'no' otherwise #}
{{ foo ?? 'no' }}

kevinpapst: Interesting, sound like a Twig exclusive feature I didn't knew about.

src/Twig/RuntimeExtension.php
```diff
@@ -85,16 +85,12 @@ public function getMenu(Request $request): ?array
     public function getNotifications(): ?NotificationEvent
     {
         if (!$this->eventDispatcher->hasListeners(NotificationEvent::class)) {
-            return null;
+            return new NotificationEvent();
```
kevinpapst, line 88: ups, missed that, but: can this method still return null?

translations/TablerBundle.en.xlf
```diff
@@ -146,6 +146,18 @@
...
+          <target>Show notifications</target>
+      </trans-unit>
+      <trans-unit id="F5Dfzd" resname="You have no notification">
+          <source>You have no notification</source>
+          <target>You have no notification</target>
+      </trans-unit>
+      <trans-unit id="Pf48Zd" resname="And %count% more…">
+          <source>And %count% more…</source>
+          <target>And %count% more…</target>
+      </trans-unit>
```
kevinpapst, line 160: For German:
```xml
      <trans-unit id="465zeJ" resname="Show notifications">
          <source>Show notifications</source>
          <target>Zeige Benachrichtigungen</target>
      </trans-unit>
      <trans-unit id="F5Dfzd" resname="You have no notification">
          <source>You have no notification</source>
          <target>Es gibt keine Benachrichtigungen</target>
      </trans-unit>
      <trans-unit id="Pf48Zd" resname="And %count% more…">
          <source>And %count% more…</source>
          <target>Und %count% weitere…</target>
      </trans-unit>
```

src/Twig/RuntimeExtension.php
cavasinf: As `{% if notifications is not null %}` was removed form the template, **NO**.
And you asked for that: https://github.com/kevinpapst/TablerBundle/pull/176#discussion_r1381603445 😄

kevinpapst: Hahaha, and are you going to change it?

kevinpapst: I am talking about

```php
    public function getNotifications(): ?NotificationEvent
```

vs

```php
    public function getNotifications(): NotificationEvent
```

cavasinf: Oh yes sure!
(didn't see you edited message above)

kevinpapst (approved):

kevinpapst:
And now I almost missed the CI Linting failures. Can you run `bon/console codestyle-fix`?

cavasinf:
> And now I almost missed the CI Linting failures. Can you run `bon/console codestyle-fix`?

Done 👍  
Phpstan too

kevinpapst (approved):

kevinpapst:
Thanks 😄

PR #177, opened 2023-11-09
[Status] Add `attr` option + deprecated non camelCase option

cavasinf:
## Description

- Deprecate non camelCase `with_dot` option
- Add safe `bool` check on options values
- ~Add `status_dot` component~
- add `attr` option to each component

### Usage
```twig
<div>
    {{ status('Not animated', {withDot : true, animated : false, color: 'blue'}) }}
    {{ status('Lite with dot', {withDot : true, color: 'red', lite: true}) }}
    {{ status('No dot', {color: 'green'}) }}
    {{ status('id', {attr: {id : 'status_id'}}) }}

    {{ status_dot() }}
    {{ status_dot({color : 'purple'}) }}
    {{ status_dot({color : 'red', animated : false}) }}
    {{ status_dot({attr: {id : 'status_dot_id'}}) }}

    {{ status_indicator() }}
    {{ status_indicator({color : 'orange'}) }}
    {{ status_indicator({color : 'red', animated : false}) }}
    {{ status_indicator({attr: {id : 'status_indicator_id'}}) }}
</div>
```

### Visual

status

## Types of changes

## Checklist

### DIFF
--- a/templates/components/status.html.twig
+++ b/templates/components/status.html.twig
@@ -20,7 +20,7 @@
         {{ utils.attr_to_html(_attr) }}
     >
         {% if _withDot %}
-            {{ status_dot({animated : _animated}) }}
+            {{ status_dot({color: _color, animated : _animated}) }}
         {% endif %}
         {{ text }}
     </span>
--- a/templates/components/status.html.twig
+++ b/templates/components/status.html.twig
@@ -14,17 +14,15 @@
     {% set _extraClass = options.extraClass ?? '' %}
     {% set _attr = options.attr ?? {} %}
 
-    {%- apply spaceless -%}
-        <span
-            class="status status-{{ _color }} {% if _lite %}status-lite{% endif %} {{ _extraClass }}"
-            {{ utils.attr_to_html(_attr) }}
-        >
-            {% if _withDot %}
-                {{ _self.status_dot({animated : _animated}) }}
-            {% endif %}
-            {{ text }}
-        </span>
-    {%- endapply -%}
+    <span
+        class="status status-{{ _color }} {% if _lite %}status-lite{% endif %} {{ _extraClass }}"
+        {{ utils.attr_to_html(_attr) }}
+    >
+        {% if _withDot %}
+            {{ _self.status_dot({animated : _animated}) }}
+        {% endif %}
+        {{ text }}
+    </span>
 {% endmacro %}
 
 {% macro status_dot(options) %}
@@ -35,10 +33,8 @@
     {% set _extraClass = options.extraClass ?? '' %}
     {% set _attr = options.attr ?? {} %}
 
-    {%- apply spaceless -%}
-        <span
-            class="status-dot {% if _animated %}status-dot-animated{% endif %} {% if _color is not null %}status-{{ _color }}{% endif %} {{ _extraClass }}"
-            {{ utils.attr_to_html(_attr) }}
-        ></span>
-    {%- endapply -%}
+    <span
+        class="status-dot {% if _animated %}status-dot-animated{% endif %} {% if _color is not null %}status-{{ _color }}{% endif %} {{ _extraClass }}"
+        {{ utils.attr_to_html(_attr) }}
+    ></span>
 {% endmacro %}
--- a/templates/components/status.html.twig
+++ b/templates/components/status.html.twig
@@ -1,5 +1,6 @@
 {% macro status(text, options) %}
     {% import '@Tabler/includes/utils.html.twig' as utils %}
+    {% import '@Tabler/components/status_dot.html.twig' as status_dot %}
 
     {% set _color = options.color ?? 'green' %}
     {% set _animated = (options.animated ?? true) is same as true %}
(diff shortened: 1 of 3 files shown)


templates/components/status.html.twig
```diff
@@ -1,16 +1,45 @@
...
-        <span class="status status-{{ _color }} {{ _lite ? 'status-lite' : '' }} {{ _extraClass }}">
-        {% if _with_dot %}
-            <span class="status-dot {{ _animated ? 'status-dot-animated' : '' }}"></span>
-        {% endif %}
-            {{ text }}
-        </span>
-    {%- endapply -%}
-{% endmacro %}
+{% macro status(text, options) %}
+    {% import '@Tabler/components/status.html.twig' as macro %}
```
cavasinf, line 2: We can maybe rename `macro` to `self`, because macro is already a name in twig.
```diff
-    {% import '@Tabler/components/status.html.twig' as macro %}
+    {% import '@Tabler/components/status.html.twig' as self %}

```

kevinpapst: Not needed: you can always reference macros in the same file via `_self.status_dot()`

cavasinf: I thought about it.
But the fact that devs will call the macro with:
```twig
{% from'@Tabler/components/status.html.twig' import status %}
```
I don't know if `_self` in that macro will be:
1. The devs current twig context 
2. OR the Bundle component context.

To be safe, I've done it hits way.

templates/components/status.html.twig
```diff
@@ -1,16 +1,45 @@
...
+    {% set _extraClass = options.extraClass ?? '' %}
+    {% set _attr = options.attr ?? {} %}
+
+    {%- apply spaceless -%}
+        <span
+            class="status status-{{ _color }} {% if _lite %}status-lite{% endif %} {{ _extraClass }}"
+            {{ utils.attr_to_html(_attr) }}
+        >
+            {% if _withDot %}
+                {{ macro.status_dot({animated : _animated}) }}
```
kevinpapst, line 24: Lets clean that up, `_self` always works:
https://twig.symfony.com/doc/3.x/tags/macro.html#importing-macros

templates/components/status.html.twig
```diff
@@ -1,16 +1,45 @@
...
+
+{% macro status_dot(options) %}
+    {% import '@Tabler/includes/utils.html.twig' as utils %}
+
+    {% set _color = options.color ?? null %}
+    {% set _animated = (options.animated ?? true) is same as true %}
+    {% set _extraClass = options.extraClass ?? '' %}
+    {% set _attr = options.attr ?? {} %}
+
+    {%- apply spaceless -%}
```
kevinpapst, line 39: You know that Twig itself warns about the `spaceless` usage?
https://twig.symfony.com/doc/3.x/filters/spaceless.html

I would remove it.

templates/components/status.html.twig
cavasinf: I don't know why, but I have something somewhere in my brain that told me to be careful with that when used inside embed or other complex import..

I'll change that.

templates/components/status.html.twig
cavasinf: No ?? Why ? I can't find anything about warnings in the doc 🤔 

How are we supposed to do? Put hyphens at every twig functions usages?

kevinpapst: Bildschirmfoto 2023-12-11 um 12 16 57

I would only use the `{{- -}}` syntax if needed

templates/components/status.html.twig
kevinpapst: And I remember that using _self wasn't possible at all ... We are doing this job for too long 😁

kevinpapst:
Do we have demos for these elements? I know you don't need them, but its really nice for visual review purposes.

kevinpapst (dismissed):

cavasinf:
Looking through TablerBundle-Demo, I've found that I am blind..
There's already a `status_dot` component..

And even more, there's already demos for those components: https://github.com/kevinpapst/TablerBundle-Demo/blob/main/templates/components/status/status.html.twig

(happy ending 🤯)

cavasinf:
> and you committed it last time

Yes, I (re)did it in the `status.html.twig` component.

> I told you: we are doing this job for too long

Christmas is almost here 😫

templates/components/status.html.twig
```diff
@@ -1,16 +1,27 @@
...
-        {% if _with_dot %}
-            <span class="status-dot {{ _animated ? 'status-dot-animated' : '' }}"></span>
-        {% endif %}
-            {{ text }}
-        </span>
-    {%- endapply -%}
-{% endmacro %}
+{% macro status(text, options) %}
+    {% import '@Tabler/includes/utils.html.twig' as utils %}
+    {% import '@Tabler/components/status_dot.html.twig' as status_dot %}
```
cavasinf, line 3: Is this ok to import only inside a `if`?

Like that:
```twig
{% if _withDot %}
   {% import '@Tabler/components/status_dot.html.twig' as status_dot %}
   {{ status_dot({color: _color, animated : _animated}) }}
{% endif %}
```

Or it is too much?

kevinpapst: Not worth it IMO. 
I am not entirely sure, but I believe it will be imported anyway during compilation.

templates/components/status_indicator.html.twig
```diff
@@ -1,13 +1,17 @@
...
+{% macro status_indicator(options) %}
+    {% import '@Tabler/includes/utils.html.twig' as utils %}
+
+    {% set _color = options.color ?? 'green' %}
+    {% set _animated = (options.animated ?? true) is same as true %}
+    {% set _extraClass = options.extraClass ?? '' %}
+    {% set _attr = options.attr ?? {} %}
+
+    <span
+        class="status-indicator {% if _animated %}status-indicator-animated{% endif %} {{ 'status-'~_color }} {{ _extraClass }}"
```
kevinpapst, line 10:
```suggestion
        class="status-indicator {% if _animated %}status-indicator-animated{% endif %} status-{{ _color }} {{ _extraClass }}"

```

kevinpapst (approved):

kevinpapst:
Merci beaucoup 👍

cavasinf:
Mit Vergnügen 😄

PR #180, opened 2023-11-20
[Vertical layout] Remove `data-bs-theme` from header

cavasinf:
## Description
This fixes code/logic that is not in Tabler today. 
See `<header>` at https://preview.tabler.io/layout-combo.html

It ends up as a bug when you want to redefine the primary color:

As it is today:
image

-------

How it should be (as Tabler demo):
image

Works as expected in mobile view:
image

## Types of changes

## Checklist

### DIFF
--- a/templates/layout-vertical.html.twig
+++ b/templates/layout-vertical.html.twig
@@ -55,7 +55,7 @@ Enjoy your theme!
 
     {% block header %}
         {% if not tabler_bundle.isCondensedNavbar() %}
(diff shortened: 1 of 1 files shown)


cavasinf:
Is this considered as a BC break? 🤔

kevinpapst:
My app uses default colors and dark/light switch.
Removing the attribute doesn't change anything here.
And if it fixes a bug, I doubt that this is a BC break. 

Did you check horizontal?
The tabler demo doesn't have it there as well, but our horizontal theme still has it:
https://github.com/kevinpapst/TablerBundle/blob/main/templates/layout-horizontal.html.twig#L30

kevinpapst:
Oh wow, how annoying 😁 do you need a new minor release?

kevinpapst (approved):

kevinpapst:
Thanks @cavasinf 👍

PR #195, opened 2024-04-26
first draft for ux icons support

### LINKED ISSUE #194: [Icon] Bye bye `tabler_icon` ➡️  Welcome Symfony UX Icon 🎉
As already discussed in #17, we've got a gap in how to handle icons in this bundle:
- Lib Tabler icons? not really (as it should because of the theme)
- Lib FontAwesome icon (which version? How to switch to Tabler? ...)
- `<svg>`? `<i>`?
- Tabler's `'icon' class` that doesn't work at all with FontAwesome

## Possible solution
We've been using [Symfony Ux icon](https://ux.symfony.com/icons) for the past months and it's been a total game-changer! It works like a charm and the best part is how easy it is to switch between icon sets (Bootstrap, Tabler, FontAwesome).
And we can send the [search page](https://ux.symfony.com/icons) to our clients and let them find what they like on their own.

## Great feature of today's bundle icons
The best part of `tabler_icon` today is the `shortcut/library` of words that are equal to an icon:
```yaml
tabler:
   icons:
      issue: fas fa-bug
```

### Same concept with UX Icon
But we ended to reproduce this same `feature` with UX Icon.
1. By ["importing" (downloading)](https://symfony.com/bundles/ux-icons/current/index.html#downloading-icons) the icon:
    ```bash
      php bin/console ux:icons:import tabler:bug
    ```
    Downloaded at: `assets/icons/tabler/bug.svg`

2. Rename it to our "shortcut" and move it to the parent icon folder:
   `assets/icons/issue.svg`

3. We can now call it like this:
    - `{{ ux_icon('issue') }}`
    - Or with twig component: `<twig:ux:icon name="issue" />`

## TablerBundle pros
With that external bundle, we can get out of a tight spot with "custom" code that isn't related to the Tabler theme
Plus, we get to remove code to maintain! 

## TablerBundle cons
- Deprecating `tabler_icon` 😢 

WDYT ?

kevinpapst:
## Description

This is a first draft for UX icons integration.
It supports both Fontawesome webfont and UX icons.

It ships a command that can convert current webfont configuration and download all icons configured at `tabler.icons`, and it even supports converting old FA5 to FA6 names (the webfont supported both versions, iconify only the new names).

The `tabler_icon()` Twig functions works as before and supports both implementations.
The `|tabler_icon` Twig filter will still work with webfonts, but not with UX icons.

The biggest issues I see:
- It increases the page size by 20kB for each request
- And many icons [cannot be cached](https://symfony.com/bundles/ux-icons/current/index.html#caching) e.g. the entire menu is based on runtime events and those icons won't be cached (according to the docs). That means for every page load PHP will read all these SVGs from disk? If that is true (I need to investigate further) this is a **blocker** for me as of now.

Closes #194 

## Types of changes

## Checklist
- [ ] I agree that this code will be published under the [MIT license](https://github.com/kevinpapst/TablerBundle/blob/master/LICENSE)


cavasinf:
Big 👍 for splitting `RuntimeExtension` in `IconExtension`.

config/packages/tabler.yaml
```diff
@@ -44,35 +44,34 @@ tabler:
...
     icons:
         # used for the action_collapsebutton components
-        collapse: fas fa-chevron-down
+        collapse: "fa6-solid:chevron-down"
         # used for the password-reset template
-        mail: far fa-envelope
+        mail: "fa6-regular:envelope"
         # used for the 404/500/... error page
-        back: fas fa-long-arrow-alt-left
+        back: "fa6-solid:left-long"
```
cavasinf, line 51: Do we still keep the `icons` config part in the future, or just during the temporary "deprecated" period?

kevinpapst: Keep it during the deprecation period. I want to try to make this PR completely BC safe.

kevinpapst: On the other hand, this is a tiny layer on top of the UX icons later, that helps managing and downloading them. 
Not sure yet ...

src/Twig/IconExtension.php
```diff
@@ -0,0 +1,52 @@
...
+
+    public function createIcon(string $name, bool $withIconClass = false, string $default = null): string
+    {
+        $icon = $this->getIcon($name, $default);
+
+        if (str_contains($icon, ':')) {
+            return $this->iconRenderer->renderIcon($icon, ['class' => 'icon']);
+        }
+
+        @trigger_error('Support for webfonts is deprecated. Switch to UX icons. Using twig function tabler_icon("' . $name . '") is deprecated.', \E_USER_DEPRECATED);
```
kevinpapst, line 34: Using twig function tabler_icon("' . $name . '") without an iconify identifier is deprecated

kevinpapst, line 34:
```suggestion
        @trigger_error('Using twig function tabler_icon("' . $name . '") without an iconify identifier is deprecated, see https://github.com/kevinpapst/TablerBundle/pull/195.', \E_USER_DEPRECATED);
```

kevinpapst:
I haven't worked on that since creating the PR, so I don't know.
If we create a v2, w should investigate if and if yes, how far we want to integrate with Symfony UX

PR #201, opened 2024-07-19
Fluent models

3lm4r:
## Description
The pull request add the Fluent pattern to the basic models. It should break nothing because void return types are exchenged to self. There can't be anybody how is checking his return types to void. Makes no sense. Fluent gives only a add.

## Types of changes

## Checklist


kevinpapst (commented):
Thanks for taking the time to contribute!

You are breaking the contract of all model interfaces.
Minimal benefit, but the necessity for a new mayor release: I will not merge that change.

3lm4r:
I'm sorry .. didn't watch the Interfaces. You are right it's dead with that. I hate non fluance on "mass production" like the menu items so i was a little too enthusastic.

sry for the inconvinience.

PR #210, opened 2025-01-24
[Login] Show password in login/sign up form

### LINKED ISSUE #209: [Login] Show password in login/sign up form
One of our clients has requested the ability to see their password while typing during login.
We can implement this on our side, but I was considering adding it directly into the template.

Looking at the Tabler demo (https://preview.tabler.io/sign-in.html), there is an icon to toggle password visibility, but no JS is attached to it. 
image

We could do the same a their, but include a simple JS embedded in the page (avoiding the need to build a separate script).

We could make this feature configurable through the TableBundle by adding a parameter like "login_show_password", defaulting to `true`.
This would allow flexibility if someone wants to disable the feature in the future.

WDYT @kevinpapst ?

cavasinf:
## Description
Fix #209 

Login

## Types of changes

## Checklist

### DIFF
--- a/templates/security.html.twig
+++ b/templates/security.html.twig
@@ -83,34 +83,36 @@
                                 <script>
                                     function toggleShowPasswordInput() {
                                         const input = document.querySelector('input[name="_password"]');
-                                        if (!input) return;
+                                        if (!input) {
+                                            return;
+                                        }
 
                                         const a = input.parentElement.querySelector('.input-group-text a');
-                                        let tooltipEl;
+                                        let tooltipEl = null;
                                         if (a.hasAttribute('aria-describedBy')) {
                                             tooltipEl = document.getElementById(a.getAttribute('aria-describedBy'));
                                         }
                                         const svgs = input.parentElement.querySelectorAll('.input-group-text svg');
 
-                                        let title
+                                        let title = '';
                                         if (input.type === 'password') {
-                                            // eye-off
                                             title = a.getAttribute('data-hide-title');
                                             input.type = 'text';
                                             svgs[0].parentElement.hidden = true;
                                             svgs[1].parentElement.hidden = false;
                                         } else {
-                                            // eye
                                             title = a.getAttribute('data-show-title');
                                             input.type = 'password';
                                             svgs[0].parentElement.hidden = false;
                                             svgs[1].parentElement.hidden = true;
                                         }
 
-                                        // Tooltip
                                         a.ariaLabel = title;
+                                        a.title = title;
                                         a.setAttribute('data-bs-original-title', title);
-                                        if (tooltipEl) tooltipEl.querySelector('.tooltip-inner').innerText = title;
+                                        if (tooltipEl !== null) {
+                                            tooltipEl.querySelector('.tooltip-inner').innerText = title;
+                                        }
                                     }
                                 </script>
                                 {% endblock %}
--- a/templates/security.html.twig
+++ b/templates/security.html.twig
@@ -140,18 +140,18 @@
 
     <script>
         function toggleShowPasswordInput() {
-            const input = document.querySelector('input[name="_password"]')
-            if (!input) return
+            const input = document.querySelector('input[name="_password"]');
+            if (!input) return;
 
-            const svgs = input.parentElement.querySelectorAll('.input-group-text svg')
+            const svgs = input.parentElement.querySelectorAll('.input-group-text svg');
             if (input.type === 'password') {
-                input.type = 'text'
-                svgs[0].parentElement.hidden = true
-                svgs[1].parentElement.hidden = false
+                input.type = 'text';
+                svgs[0].parentElement.hidden = true;
+                svgs[1].parentElement.hidden = false;
             } else {
-                input.type = 'password'
-                svgs[0].parentElement.hidden = false
-                svgs[1].parentElement.hidden = true
+                input.type = 'password';
+                svgs[0].parentElement.hidden = false;
+                svgs[1].parentElement.hidden = true;
             }
         }
     </script>
--- a/templates/security.html.twig
+++ b/templates/security.html.twig
@@ -25,98 +25,98 @@
 
         <div class="{% block card_classes %}card card-md{% endblock %}">
             {% block login_box %}
-                <div class="card-body">
-                    <h2 class="card-title text-center mb-4">{% block login_box_msg %}{% endblock %}</h2>
-                    {% block login_box_error %}
-                        {% if error|default(false) %}
-                            <div class="alert alert-important alert-danger">{{ error.messageKey|trans(error.messageData, 'security') }}</div>
-                        {% endif %}
-                    {% endblock %}
-                    {% block login_form %}
-                        <form action="{{ path('tabler_login_check'|tabler_route) }}" method="post" autocomplete="off" class="login-box-body security-login">
-                            {% block login_form_start %}{% endblock %}
-                            <div class="mb-3">
-                                <label for="username" class="form-label">{{ 'Username'|trans({}, 'TablerBundle') }}</label>
-                                <input autocomplete="username" type="text" id="username" name="_username" tabindex="10" class="form-control" placeholder="{{ 'Username'|trans({}, 'TablerBundle') }}" value="{{ last_username|default('') }}" required="required">
-                            </div>
-                            <div class="mb-3">
-                                <label for="password" class="form-label">
-                                    {{ 'Password'|trans({}, 'TablerBundle') }}
-                                    {% block password_forgotten %}
-                                        {% if 'tabler_password_reset'|tabler_route != 'tabler_password_reset' %}
-                                            <span class="form-label-description">
+            <div class="card-body">
+                <h2 class="card-title text-center mb-4">{% block login_box_msg %}{% endblock %}</h2>
+                {% block login_box_error %}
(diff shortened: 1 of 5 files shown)


cavasinf:
A new translation, `Show password`, has been added.  
The feature has been implemented within a `block` instead of as an option, to simplify the structure.

templates/security.html.twig
```diff
@@ -52,6 +52,33 @@
...
                             <div class="input-group input-group-flat">
                                 <input autocomplete="new-password" id="password" name="_password" type="password" tabindex="20" class="form-control" placeholder="{{ 'Password'|trans({}, 'TablerBundle') }}" required="required">
+                                {% block password_show %}
+                                <span class="input-group-text">
+                                    <a
+                                            href="#" class="link-secondary"
+                                            data-bs-toggle="tooltip" title="{{ 'Show password'|trans({}, 'TablerBundle') }}"
+                                            onclick='toggleShowPasswordInput()'
+                                    >
+                                        <div>
```
kevinpapst, line 62: Block element `div` inside inline-element `span` is invalid HTML

templates/security.html.twig
```diff
@@ -52,6 +52,33 @@
...
+                                            <!-- Download SVG icon from http://tabler.io/icons/icon/eye -->
```
kevinpapst, line 63: lets remove these comments from HTML output, replace with twig comment `{# #}`

templates/security.html.twig
```diff
@@ -52,6 +52,33 @@
...
+                                            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="icon icon-1">
+                                               <path d="M10 12a2 2 0 1 0 4 0a2 2 0 0 0 -4 0"></path>
+                                               <path d="M21 12c-2.4 4 -5.4 6 -9 6c-3.6 0 -6.6 -2 -9 -6c2.4 -4 5.4 -6 9 -6c3.6 0 6.6 2 9 6"></path>
+                                            </svg>
+                                        </div>
+
+                                        <div hidden>
```
kevinpapst, line 70: There is a `hidden` attribute since 10+ years and I never heard about it? I can't believe it.... thanks for using it here 👍  will use it in the future.

kevinpapst (changes requested):

templates/security.html.twig
cavasinf: This what the JS `hidden` property changes 😉 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden

templates/security.html.twig
```diff
@@ -110,6 +137,24 @@
 
 {% block javascripts %}
     <script src="{{ asset('bundles/tabler/tabler.js') }}?{{ tabler_asset_version() }}"></script>
+
+    <script>
+        function toggleShowPasswordInput() {
+            const input = document.querySelector('input[name="_password"]')
```
kevinpapst, line 143: Okay, I am sorry about that code style nit-picking, but I missed it before.

Please add `;`. 
I not they are not necessary in JS world, but I already ran into bugs due to missing semicolons. 

I assume you have tested your changes, so I can safely merge?!

cavasinf: > Okay, I am sorry about that code style nit-picking, but I missed it before.
> 
> Please add ;. I know that they are not necessary in JS world, but I already ran into bugs in the past due to missing semicolons.

It's been a long day since I wrote semicolons for JS/TS, but no problem 😃 

> I assume you have tested your changes, so I can safely merge after that one change?!

Yep 👍 

Login

cavasinf:
Wait.. should we add at least the `en` trans for "Show password" or weblate will find it magically?

kevinpapst:
Yes, good catch! We need the translation in the base language, so Weblate can pick it up.

cavasinf:
Hey, busy weeks lately, but we have basic trans now (+ fr)

kevinpapst (approved):

templates/security.html.twig
```diff
@@ -52,6 +52,46 @@
...
+                                            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="icon icon-tabler icons-tabler-outline icon-tabler-eye-off">
+                                                <path stroke="none" d="M0 0h24v24H0z" fill="none" />
+                                                <path d="M10.585 10.587a2 2 0 0 0 2.829 2.828" />
+                                                <path d="M16.681 16.673a8.717 8.717 0 0 1 -4.681 1.327c-3.6 0 -6.6 -2 -9 -6c1.272 -2.12 2.712 -3.678 4.32 -4.674m2.86 -1.146a9.055 9.055 0 0 1 1.82 -.18c3.6 0 6.6 2 9 6c-.666 1.11 -1.379 2.067 -2.138 2.87" />
+                                                <path d="M3 3l18 18" />
+                                            </svg>
+                                        </span>
+                                    </a>
+                                </span>
+                                <script>
```
kevinpapst, line 77: when i tried it in my project, the javascript was missing, because I use my own build and have overwritten the `javascript` block. even though its not best practice, i think we can inline that in this case.

cavasinf:
> To me the open eye says "I can see the password" and the strikethrough eye means "password cannot be seen".
> Don't you think the eyes are in the wrong order?

I understand your vision of the button.  
The button is not meant to indicate the **current state** of the input but rather the **consequence of clicking it**.  

After looking at some references, here are a few examples:  
- [GitLab](https://gitlab.com/users/sign_in)  
- [OpenAI](https://openai.com/)  
- [Netflix](https://www.netflix.com/login)  
- [CNN](https://edition.cnn.com/account/log-in)  
- [Odoo](https://www.odoo.com/de_DE/web/login)  

Considering your perspective, I believe we should add a new label: **"Hide password"**.  
This emphasizes the **action** rather than the **state**.

templates/security.html.twig
cavasinf: Is this not an expected behavior?  
If you **override** the whole JS, it is expected to **lose** this function—unless you add `{{ parent() }}` in the block.  
But in that case, it is no longer a full override, rather more like an extension.

kevinpapst: For me that is not expected. 
Being able to replace the build by replacing the entire javascript block is one thing, but loosing the feature seems contra intuitive to me.

kevinpapst:
Ok, if these platforms have it this way, I am fine with it as well.
But regarding the translation, you are right. 
Having "Show password" there, no matter which state, is misleading/wrong. Shall we remove it altogether or do you want to add a second translation?

templates/security.html.twig
cavasinf: Since the code has a safe exit (checks if input exists), we can force the to template by extracting it from the `javascript` block.
If devs use the security template without the `_password` input, the js will still be there, but will do nothing.
But this means that devs cannot get rid of the js we provide.

cavasinf:
> Shall we remove it altogether or do you want to add a second translation?

I'll add the second translation, and the js will only hide/show the `a` element, which will be doubled with the two different titles. (no need to target the svgs anymore)

templates/security.html.twig
kevinpapst: I don't get that. How can you remove the password field without the javascript in the current version? 
If you see the eye icons, you get the javascript. Otherwise you don't.

kevinpapst:
I like the SVG solution in this case.

cavasinf:
Ok, a little complication with this simple feature. Since the tooltip is a `document` level element, we cannot simply hide/show the `a` element with the title, because if the tooltip is already displayed, nothing will change.
Instead, we need to change the `ariaLabel` + `data-bs-original-title` attribute + the currently displayed tooltip.

hide-show

kevinpapst:
Hahaha, and that's why customer tend to say "but its only" ... without understanding all the problems that can arise from "only a link" 😁 

Its is ready now from your end?

cavasinf:
> Hahaha, and that's why customer tend to say "but its only" ... without understanding all the problems that can arise from "only a link" 😁

Yeah, every time I open a PR, we always exchange with 15+ comments on it 😭 
So I get the felling 😉 

> Its is ready now from your end?

- You can maybe add the `de` trans
- How do you generate the `id` prop for the trans? I saw you did it in your commit.

kevinpapst (approved):

kevinpapst:
The `title` attribute was still missing, I added the translations and the ids, which I generate with this command:
https://github.com/kimai/kimai/blob/main/src/Command/TranslationCommand.php

The command has many features which I need to manage translation files.

If you have no objections, I will merge and release

cavasinf:
> The title attribute was still missing

Yeah, I thought as long as the bootstrap tooltip is init, `title` is not really mandatory anymore since bootstrap will never read title again when initialized. `aria-label` was more important as it is used by screen readers.
But useful if one day bootstrap doesn't work anymore, the native browser tooltip will works.

> The command has many features which I need to manage translation files.

Nice 👍 

> If you have no objections, I will merge and release

No objections 🫡

kevinpapst:
I don't see the bootstrap title, as I use `data-toggle` instead of `data-bs-toggle`, which is obviously my own problem 😁 but that's why I immediately saw it.

PR #212, opened 2025-02-18
[Doc] Update customize doc as bootstrap ways

cavasinf:
## Description
Update customize doc as bootstrap ways

## Types of changes

## Checklist


kevinpapst:
This version sounds better. 

I wonder how much this might change with the upcoming release of:
https://tabler-git-dev-props-tabler-io.vercel.app/?theme-primary=red&theme-font=monospace&theme-radius=2&theme-base=neutral

CSS variables might become the simpler solution, so people don't need the additional build step

cavasinf:
> CSS variables might become the simpler solution, so people don't need the additional build step

Agree, even more with the [Symfony Asset Mapper](https://symfony.com/doc/current/frontend/asset_mapper.html), which encourages a "no-build" approach (no webpack) using raw JS and CSS, without the need for TS or SCSS files.

kevinpapst (approved):
Thanks 👍

PR #214, opened 2025-02-25
Remove `.row` and `.row-cards` classes from container + remove `section` element

### LINKED ISSUE #213: [Layout] Too many "forced" content for inner `.page-body`
Hey,
We’re currently a bit stuck with how the `layout-xxx.html.twig` file define the base 'structure' of the page.

## Objective
The goal is to create a dashboard page with multiple rows and columns of with **the same height**.  

## Current state
Currently, the base template forces this HTML structure:

```html
<div class="row row-cards">          
   <section id="" class="content">
       [...]
   </section>
</div>
```
https://github.com/kevinpapst/TablerBundle/blob/39b5e2946c0d4fea5ad05be9b1cdad686d7035b3/templates/layout-vertical.html.twig#L120-L128

### Current render
Here's what we have currently:
Image

### Desired render:
Image

To achieve this, we need to apply the `.row-deck` class directly after the top-level `.container` class.  
But we want the `.row-deck` in the top div + remove the section part.

The whole HTML wanted:
Image

## Question/Solution

1. Is there specific reason for having the `section` block + default `row row-cards` classes?
2. It may be a relic of `admin-lte` bundle?

It might be more flexible if this part of the structure was left for the developers to define, rather than enforced by the top-level template.

References with differents classes just after the `.container` div:
https://preview.tabler.io/
https://preview.tabler.io/activity.html
https://preview.tabler.io/photogrid.html

cavasinf:
## Description
Fix #213 

## Types of changes

## Checklist

### DIFF
--- a/templates/layout-horizontal.html.twig
+++ b/templates/layout-horizontal.html.twig
@@ -122,17 +122,13 @@ Enjoy your theme!
 
         <div class="page-body">
             <div class="{{ ''|tabler_container }}">
-                <div class="row row-cards">
-                    {% block page_content_before %}{% endblock %}
+                {% block page_content_before %}{% endblock %}
 
-                    <section id="{% block page_content_id %}{% endblock %}" class="{% block page_content_class %}content{% endblock %}">
-                        {% block page_content_start %}{{ include('@Tabler/includes/flash_messages.html.twig') }}{% endblock %}
-                        {% block page_content %}{% endblock %}
-                        {% block page_content_end %}{% endblock %}
-                    </section>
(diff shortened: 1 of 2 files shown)


kevinpapst (dismissed):
I tested these changes and removing the `<div>` was no problem. But I cannot work without the section (which would require a lot of refactoring) 🤷 

So I did as suggested:
```twig
{% block page_content_before %}
    <section class="content {% block page_class %}{% endblock %}">
{% endblock %}
{% block page_content_after %}
    </section>
{% endblock %}
```
We should document that in a UPGRADING.md.

kevinpapst (approved):

PR #216, opened 2025-03-06
[Component] Button - Change handled disabled on `a` type

cavasinf:
## Description
Change the disabled effect on a link by following the [W3 recommendation](https://www.w3.org/TR/html-aria/#el-a)

## Types of changes

## Checklist


cavasinf:
Since the `href` attribute is conditionally removed, I've put it in `BC Break`.

PR #218, opened 2025-04-15
Fix Twig 3.15 deprecation on ?? usage

thePanz:
Fix the deprecation `Deprecated: Since twig/twig 3.15: Add explicit parentheses around the "??" binary operator to avoid behavior change in the next major version as its precedence will change in "@Tabler/components/dropdown.html.twig" at line 29. `

## Types of changes

## Checklist


kevinpapst (approved):
Thanks 👍

PR #220, opened 2025-05-16
Use non-deprecated Extension class

thePanz:
## Description
This solves the deprecation reported when running with Symfony v7.2:

> The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "KevinPapst\TablerBundle\DependencyInjection\TablerExtension".

## Types of changes

## Checklist

See as an example: https://github.com/hwi/HWIOAuthBundle/pull/2005

### DIFF
--- a/src/DependencyInjection/TablerExtension.php
+++ b/src/DependencyInjection/TablerExtension.php
@@ -11,9 +11,9 @@ namespace KevinPapst\TablerBundle\DependencyInjection;
 
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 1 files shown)


kevinpapst (approved):
Works in 6.4 as well. The class was [existing in 6.0](https://github.com/symfony/symfony/blob/6.0/src/Symfony/Component/HttpKernel/DependencyInjection/Extension.php) already - so can be merged safely. 
Thanks 👍

PR #230, opened 2025-11-17
Tabler 1.4

kevinpapst:
## Description

**Before:**
- Tabler v1.0.0-beta21 
- Bootstrap v5.3.3
- Fontawesome 6.5

**After:**
- Tabler v1.4.0 
- Bootstrap v5.3.7
- Fontawesome 6.7.2

See https://github.com/kevinpapst/TablerBundle-Demo/pull/25

New config options available (also available as getter/setter in `ContextHelper`):
```yaml
tabler:
    options:
        # slate, gray, zinc, neutral, stone
        theme_base: "slate"
        # possible values: 0, 0.5, 1, 1.5, 2 | radius in pixels * 4
        theme_radius: "0.5"
        # blue, azure, indigo, purple, pink, red, orange, yellow, lime, green, teal, cyan
        theme_primary: "blue"
```

Introduced a new `_tabler-fixes.scss` file to have a place to add bugfixes for known issues in Tabler itself.

You need to pay attention to:
- `badges` (add a class for the text color)
- the `data-bs-theme` attribute was moved from `<body>` to `<html>`
- new `data-bs-theme-*` attributes were added  to `<html>`

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
- [ ] I updated the documentation (see [here](https://github.com/kevinpapst/TablerBundle/tree/master/docs))
- [ ] I agree that this code will be published under the [MIT license](https://github.com/kevinpapst/TablerBundle/blob/master/LICENSE)

PR #264, opened 2026-09-10
Fix the page header text color with an overlapping navbar

### LINKED ISSUE #243: Page title not styled correctly when using "header_dark" and "navbar_overlap"
Since upgrading to version 2 we noticed an incorrect styling for page titles making them hardly readable.

I could reproduce it by creating a new test project using a base.html.twig template containing only:
```
{% extends '@Tabler/layout_horizontal.html.twig' %}
```

In the configuration I only changed these settings:
```
header_dark: true
navbar_overlap: true
```

The result looks like:
Image

Amoifr:
## Description
Fixes #243.

Since the Tabler 1.4 update (#230), the header is `data-bs-theme="dark"` whenever `navbar_overlap` is on, and the page header sits on the dark band the navbar draws below itself. The `text-white` the page title had in 1.x was lost in that update. This puts it back, on `.page-header`, which is where Tabler's own `PageHeader` component puts it for an overlapping dark navbar.

It sits outside the `page_header_class` block, so overriding that block, as the workaround in the issue does, keeps it.

The new `LayoutHorizontalTest` renders the `page_header` block for the four combinations of `header_dark` and `navbar_overlap`, plus a block override. It fails without the template change.

## Types of changes

## Checklist


kevinpapst:
Thank you for your contribution @Amoifr 👍

PR #266, opened 2026-09-14
Add blocks around the login username and password fields

### LINKED ISSUE #258: Allow to specify username field type on login screen
The username field is defined as:
```
<input autocomplete="username" type="text" id="username" name="_username" ...
```

If the username is an email, changing the type-attribute to "email" would make the input more user-friendly.

Amoifr:
Fixes #258

`security.html.twig` now wraps the username and the password field in their own blocks, so an application can rewrite just that part of the template:

```twig
{% extends '@Tabler/security.html.twig' %}

{% block login_username %}
    <div class="mb-3">
        <label for="username" class="form-label">{{ 'Email'|trans }}</label>
        <input autocomplete="username" type="email" id="username" name="_username" tabindex="10" class="form-control" required="required">
    </div>
{% endblock %}
```

Nothing changes by default: the username input is still `type="text"`. @cavasinf is right that email would be a poor default, since plenty of projects use plain usernames or accept both, so this only makes the field replaceable.

The two blocks sit at the same level as the existing `remember_me` and `csrf_token` ones. Three tests come with it: one pins the `type="text"` default so it cannot drift, and the other two render an overriding template and check the field is really replaced. The last two fail without the template change.

Happy to expose a simple `type` variable instead, or in addition, if you would rather have that than a block override. Just say which one you prefer.


kevinpapst (approved):

kevinpapst:
Thank you @Amoifr 👍

